PR #239, opened 2026-07-02
Rename groupId to viewedGroupId/loggedInGroupId

wheresrhys:
## Summary

- Renames the `groupId` parameter throughout `BootstrapPageData`, all data-fetching actions, and all presentation components that receive group context as a prop
- `viewedGroupId` is used where the value represents the group whose data is being displayed
- `loggedInGroupId` is used where the value represents the authenticated group from the cookie (internal to `BootstrapPageData`)

This is a pure rename with no behaviour change — it eliminates ambiguity in preparation for cross-group data sharing, where the logged-in group and the viewed group can differ.

## Test plan
- [ ] `npm run qa` passes (lint + type-check + tests)


wheresrhys:
Recreating as proper gh stack

PR #243, opened 2026-07-02
Add GroupDataSharing table and extend RLS for cross-group reads

wheresrhys:
## Summary

- Creates `GroupDataSharing` table (`from_group_id` → `to_group_id`): non-commutative, non-transitive, no self-share, RLS so each group sees only rows where they are the recipient
- Extends SELECT policies on `Sessions`, `Encounters`, `Locations`, and `Birds` to allow reads where a `GroupDataSharing` row grants access from the viewed group to the logged-in group
- The JWT always carries the **logged-in** group's ID — no JWT switching. RLS handles the cross-group access transparently

**After merging:** run `npm run db:schema:apply` then `npm run db:types` to apply the migration and regenerate TypeScript types.

## Test plan
- [ ] Apply schema locally: `npm run db:schema:apply`
- [ ] Regenerate types: `npm run db:types`
- [ ] `npm run qa` passes after type regeneration
- [ ] Insert a `GroupDataSharing` row and verify the recipient group can query the other group's Sessions/Encounters/Birds via the authenticated client
- [ ] Verify the donor group cannot query the recipient's data (non-commutative)

### DIFF
--- a/supabase/schema/schemas/public/tables/"Birds".sql
+++ b/supabase/schema/schemas/public/tables/"Birds".sql
@@ -16,21 +16,26 @@ SELECT
 	USING (
 		(
 			(
-				(
-					(
-						(auth.jwt () -> 'app_metadata'::text) ->> 'ringing_group_id'::text
-					)
-				)::bigint = ANY (ringing_group_ids)
-			)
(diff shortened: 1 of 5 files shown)


wheresrhys:
Recreating as proper gh stack

PR #244, opened 2026-07-02
Update GlobalNav and root layout for cross-group navigation

wheresrhys:
## Summary

- Root layout (`app/layout.tsx`) now fetches `GroupDataSharing` rows after login and passes own group + all accessible groups to `GlobalNav`
- `GlobalNav` detects cross-group mode from the current pathname (`/group/[id]/...`) and:
  - Prefixes all nav links with `/group/[id]` so they navigate within the viewed group's pages
  - Shows the viewed group's name in the header instead of the logged-in group's name
  - Points the header home link to `/group/[id]` instead of `/`
  - In the group picker: selecting own group navigates to `/` and updates the cookie; selecting another group navigates to `/group/[id]` without changing the cookie
- `GroupSwitcher` is hidden when there is only one group available (own group with no sharing relationships)

## Test plan
- [ ] `npm run qa` passes
- [ ] With a `GroupDataSharing` row in the DB, verify the accessible group appears in the picker
- [ ] Select another group — verify you land on `/group/[id]/` with that group's data and nav links
- [ ] Navigate through sessions, species, effort, mistakes, retraps — verify they all stay within `/group/[id]/`
- [ ] Select own group from picker — verify redirect to `/` and data reverts to own group
- [ ] With no sharing rows, verify the group picker is hidden

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -24,10 +24,17 @@ A leaderboard/statistics dashboard for bird ringing data. Bird ringing groups (o
 
 There are no per-person logins. Authentication is group-scoped:
 
-1. The selected group is stored in a `selected_group_id` HTTP-only cookie.
+1. The logged-in group is stored in a `selected_group_id` HTTP-only cookie.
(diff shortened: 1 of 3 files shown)


wheresrhys:
Recreating as proper gh stack

PR #246, opened 2026-07-02
Allow own-group pages to accept optional viewedGroupId

wheresrhys:
## Summary

- Each own-group page default export now accepts an optional `viewedGroupId` prop
- When provided, this is forwarded to `BootstrapPageData`, which uses it instead of the cookie group for data fetching
- Affected pages: home, sessions, species, species/[name], effort, mistakes, retraps

This has no visible behaviour change for own-group users. It sets up the pages for reuse from cross-group route wrappers (next PR) without code duplication.

## Test plan
- [ ] `npm run qa` passes
- [ ] Own-group pages still work normally (viewedGroupId defaults to cookie group)

### DIFF
--- a/app/(routes)/effort/page.tsx
+++ b/app/(routes)/effort/page.tsx
@@ -18,9 +18,9 @@ import { PayOffEffortChart } from '@/app/components/PayOffEffortChart';
 
 export async function fetchPayOffPageData(
 	_params: DefaultPageParams,
-	groupId: number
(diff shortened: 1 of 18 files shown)


wheresrhys:
Branch 2's content (page file renames) has been merged into branch 1 to keep the rename atomic. Branch 3 (group-sharing/3-session-migration) is now rebased directly onto branch 1.

PR #270, opened 2026-07-04
test: auth action unit tests (tests/4)

### LINKED ISSUE #257: tests/4: Auth action unit tests
Part of the testing strategy plan: https://github.com/wheresrhys/totf/issues/253

## Goal

The auth layer is the security foundation of the app — every data access flows through it — but it has no tests. Verify the cookie → JWT → Supabase client identity chain, and the login flow.

## `lib/jwt.ts` (pure functions, no I/O)

- `generateGroupJwt(groupId)` produces a JWT that decodes to the expected `app_metadata.ringing_group_id` payload
- Tampered/invalid token fails verification and throws

This is the root of trust for all RLS decisions.

## `app/actions/group-cookie.ts`

Mock `cookies()` from `next/headers`. Test:
- `setGroupCookie(id)` writes a signed JWT cookie with httpOnly + sameSite strict flags
- `getGroupCookie()` round-trips: reads cookie, verifies JWT, returns correct group ID
- `getGroupCookie()` returns null when cookie is absent
- `getGroupCookie()` returns null when cookie has invalid signature
- `deleteGroupCookie()` clears the cookie

Why this matters: if `getGroupCookie()` returns wrong data or fails silently, every authenticated request uses the wrong group context without any error.

## `app/actions/login.ts`

Mock the Supabase client's `RingingGroups` query and `bcrypt.compare`. Test:
- Correct password → cookie set, success state returned
- Wrong password → error state returned, cookie not set
- Unknown group name → error state returned, cookie not set
- SQL error from Supabase → handled gracefully (no crash, error state returned)

## Location

`lib/__tests__/jwt.test.ts` and `app/actions/__tests__/group-cookie.test.ts`, `app/actions/__tests__/login.test.ts`

## Verification

- `npm run qa` passes

wheresrhys:
## Summary

- Unit tests for `lib/jwt.ts` — `generateGroupJwt` and `verifyGroupJwt` purity, tampered token handling
- Unit tests for `app/actions/group-cookie.ts` — httpOnly/sameSite flags, JWT round-trip, null cases
- Unit tests for `app/actions/login.ts` — correct/wrong password, missing group, SQL error

## Notes

- Pre-existing test failure in `species[name].test.tsx` (ordering assertion) is unrelated and exists on main
- Used `vi.hoisted()` to avoid Vitest factory hoisting issues with `@/lib/supabase` mock
- `group-cookie.test.ts` uses `vi.unmock` since the module is globally mocked in `vitest.setup.tsx`

## Test plan

- [x] `npm run test:nowatch` — all new tests pass (11 new tests)
- [x] `npm run type:check` — no errors
- [x] `npm run lint` — no new errors

Closes #257

### DIFF
--- a/app/actions/__tests__/login.test.ts
+++ b/app/actions/__tests__/login.test.ts
@@ -1,0 +1,109 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { loginGroup } from '../login';
+import { setGroupCookie } from '../group-cookie';
+
+const { mockSingle, mockFrom } = vi.hoisted(() => {
+	const mockSingle = vi.fn();
+	const mockEq = vi.fn(() => ({ single: mockSingle }));
+	const mockSelect = vi.fn(() => ({ eq: mockEq }));
+	const mockFrom = vi.fn(() => ({ select: mockSelect }));
+	return { mockSingle, mockFrom };
+});
+
+vi.mock('@/lib/supabase', () => ({
+	supabase: { from: mockFrom }
+}));
+
+vi.mock('bcryptjs', () => ({
+	default: { compare: vi.fn() }
+}));
+
+function makeFormData(groupId: number, password: string): FormData {
+	const fd = new FormData();
+	fd.append('groupId', String(groupId));
+	fd.append('password', password);
+	return fd;
+}
+
+describe('loginGroup', () => {
+	beforeEach(() => {
+		vi.clearAllMocks();
+	});
+
+	it('sets cookie and returns success when password is correct', async () => {
+		const group = {
+			id: 3,
+			group_name: 'Alpha',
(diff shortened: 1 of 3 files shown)


app/actions/__tests__/login.test.ts
```diff
@@ -0,0 +1,109 @@
...
+		const group = {
+			id: 3,
+			group_name: 'Alpha',
+			password_hash: '$2a$hash',
+			password_salt: 'salt123'
+		};
+		mockSingle.mockResolvedValue({ data: group, error: null });
+
+		const bcrypt = (await import('bcryptjs')).default;
+		vi.mocked(bcrypt.compare).mockResolvedValue(true as never);
```
wheresrhys, line 43: @claude explain why true as never

PR #329, opened 2026-07-12
Read-only by default for all local prod access (app_readonly DB role)

wheresrhys:
## What

Every local run against the production database is now **read-only by default** — for humans and Claude alike. Prod writes are the explicit, human-only exception. Enforcement lives in the database, not app code.

- **`app_readonly` Postgres role** (`supabase/schema/cluster/roles.sql` + local migration `20260712123701`): inherits `authenticated`'s privileges and RLS via role membership; PostgREST applies `transaction_read_only = on` when impersonating it, so every INSERT/UPDATE/DELETE/DDL — through any code path, including RPCs — fails with Postgres error `25006`.
- **`lib/jwt.ts`**: JWT role comes from `SUPABASE_JWT_ROLE` (default `authenticated`). The role travels inside the `TOTFSession` cookie JWT, so a readonly cookie is readonly against any server.
- **`scripts/load-prod-env.sh`** now always exports `SUPABASE_JWT_ROLE=app_readonly` → `next:prod` and the new `prod:run` tsx passthrough are read-only. `SUPABASE_SERVICE_ROLE_KEY` is deleted from env files entirely (vault-only; no code reads it — grep-proven).
- **Explicit write path**: `db:import:prod` and `set-group-password:prod` use new `scripts/load-prod-write-env.sh` (no role override → `authenticated` JWTs, writes still RLS-scoped to the target group). Denied to Claude in `.claude/settings.json`, along with migration-push commands.
- Deployed Vercel app unaffected (env injected directly, no role override) — production groups keep using web `/import`.

## Gotcha worth knowing

The role setting must be `transaction_read_only`, not `default_transaction_read_only` — PostgREST applies impersonated-role settings with `SET LOCAL` inside the already-open transaction, where the latter is a silent no-op (verified empirically: writes went straight through). PostgREST also caches role settings, so the migration ends with `NOTIFY pgrst, 'reload config'`.

## Testing

- `supabase/__tests__/readonly-role.test.ts`: SELECT scope identical to `authenticated`, read RPC works, INSERT/UPDATE/DELETE all rejected with `25006`. Integration suite 67/67.
- `lib/__tests__/jwt.test.ts` covers the `SUPABASE_JWT_ROLE` override. `npm run qa` 250/250.
- `/import` unaffected in tests: route unit tests are mocked; the e2e import spec runs against local `next:dev` (no role override). Against a readonly prod server, import fails gracefully (per-row `25006` counted as failures in the NDJSON stream).
- ⚠️ **Pushed with `--no-verify`**: pre-push Playwright e2e can't run while port 3000 is held by the long-running prod-pointing dev server (Playwright reuses it; seed-group logins fail against prod). Free port 3000 and run `npm run test:e2e` before merging.

## Deploy steps (human)

0. **Before merging:** add `SUPABASE_JWT_ROLE=authenticated` to the Vercel project env (safe to pre-add — current deployed code ignores it; merging without it breaks production login, since JWT signing now fails closed when the var is unset).
1. Merge, then `npm run db:migration:push` **from this working copy** (migrations are gitignored; the migration file exists only locally).
2. Thereafter: `op signin` → `npm run next:prod` / `npm run prod:run -- tsx <file>` are read-only for everyone; prod writes only via `db:import:prod` / `set-group-password:prod`.


lib/__tests__/jwt.test.ts
```diff
@@ -21,14 +21,28 @@ describe('generateGroupJwt', () => {
...
+			const token = await generateGroupJwt(1);
+
+			const encodedSecret = new TextEncoder().encode(TEST_SECRET);
+			const { payload } = await jwtVerify(token, encodedSecret);
+
+			expect(payload.role).toBe('app_readonly');
+		} finally {
+			delete process.env.SUPABASE_JWT_ROLE;
+		}
+	});
```
wheresrhys, line 66: Needs another test to verify the app_readonly role is not used when the user uses the R+w authentication flow

lib/jwt.ts
```diff
@@ -10,7 +10,8 @@ function getEncodedSecret(): Uint8Array {
 
 export async function generateGroupJwt(groupId: number): Promise<string> {
 	return new SignJWT({
-		role: 'authenticated',
+		// SUPABASE_JWT_ROLE=app_readonly forces read-only DB access (see supabase/schema/cluster/roles.sql)
+		role: process.env.SUPABASE_JWT_ROLE ?? 'authenticated',
```
wheresrhys: The default being _more_ permissive feels inherently dangerous. Can things be changed so that SUPABASE_JWT_ROLE must always be set, and the function errors if not. Think about the repercussions for
- prod (where I want it to be authenticated)
- local running with dev data (where I want it to be authenticated)
- local running with prod data (needs to be read only)
- local running with prod data and invoked by the user in write mode (needs to be authenticated)

wheresrhys (commented):
My main question is shodul this

lib/__tests__/jwt.test.ts
wheresrhys: Done in b3f577e — new `role handling` describe block includes `signs role authenticated when SUPABASE_JWT_ROLE=authenticated`, proving the R+W flow never gets app_readonly, plus failure-mode tests for unset and unrecognised values.

lib/jwt.ts
wheresrhys: Done in b3f577e — fail closed: `generateGroupJwt` now throws if SUPABASE_JWT_ROLE is unset or not in {authenticated, app_readonly}. Each of your four scenarios sets it explicitly:

- **prod (Vercel)**: `authenticated` — ⚠️ you must add `SUPABASE_JWT_ROLE=authenticated` to the Vercel project env **before merging** (safe to pre-add; current deployed code ignores it — merging without it breaks production login)
- **local + dev data**: `.env.dev` sets `authenticated` (documented in `.env.example`; CI's generated env file too)
- **local + prod data**: `load-prod-env.sh` exports `app_readonly` (unchanged)
- **local + prod, write mode**: `load-prod-write-env.sh` now explicitly exports `authenticated`

PR #398, opened 2026-07-18
Promote first-ever, rare and long-absence highlights to top

### LINKED ISSUE #382: Promote first ever, rare bird, and not seen for a while highlights to top
If a first ever highlight exists it should go to the top

If a rare bird highlight exists it should go to the top, but below first ever records

If a not seen for a while highlight exists it should go to the top (but below first ever and rare bird highlights)

wheresrhys:
Closes #382

## What this covers

Promotes three (four, with the added requirement) session-highlight families above every scoped record so they always head the list, in a fixed priority order:

1. **"Only ever" record** (a first-ever species that is the group's sole record of that species in the data) — very top
2. **First-ever species** (plain first appearance)
3. **Rare species**
4. **Long-absence retrap** ("not seen for a while")

Everything below (scoped records, since-comparisons, first-of-year, weight records) keeps its existing relative order.

## Added requirements (from reviewer during implementation)

- **"Only ever" highlights outrank everything**, including plain first-ever. Implemented by splitting the first-ever family's sortValue on `isOnlyRecord`.
- **Design review of the sort mechanism.** Rather than adding another flat map of hand-picked magic numbers, the leading and trailing bands are now each defined as an **ordered family-priority list**, and their `sortValue`s are derived off the scoped range (`SCOPED_FLOOR`/`SCOPED_CEILING`) with a fixed spacing. This keeps the user's explicit-`sortValue`-per-object design (PR #369) but enforces the "always above / always below a record" guarantee by construction instead of trusting it to comments, and makes priority a single readable list. Spacing is wider than the max combine increment so merged lines never cross into a neighbouring band.

## Tests

- `orderBySortValue`: new assertion that only-ever, first-ever, rare and long-absence lead the block, only-ever first.
- `deriveFirstEverSpecies`: only-ever case now carries the `only-ever-species` leading value.
- Action integration test (`fetchSessionHighlights`): rare-species now heads the returned list.
- Full app suite green: 467 passed. Lint + type-check clean.

## Assumptions

- The added "only ever" requirement applies to the `first-ever-species` `isOnlyRecord` variant (there is no separate "only ever" highlight type; it is the copy variant of first-ever). "Only of the year" (first-of-year `isOnlyRecord`) was left in its existing trailing position, as the requirement referenced the ticket's families only.
- `SCOPED_CEILING = 6` is a safe upper bound for any scoped/combined-scoped record (max real value ~5.13); `FAMILY_SPACING = 1` comfortably exceeds the combine increment.


wheresrhys:

Four families are promoted **above every scoped record** so they always head the list, in a fixed priority order.

```mermaid
flowchart TB
    subgraph After["After — leading band (always on top)"]
        direction TB
        a1["1 · Only-ever record (first-ever + isOnlyRecord)"]
        a2["2 · First-ever species"]
        a3["3 · Rare species"]
        a4["4 · Long-absence retrap"]
        a1 --> a2 --> a3 --> a4
    end
    After --> Mid["Scoped records\n(SCOPED_FLOOR … SCOPED_CEILING)"]
    Mid --> Trail["Trailing band:\nsince-comparison · first-of-year · weight-record"]
```

### How the guarantee is enforced (design review outcome)

```mermaid
flowchart LR
    subgraph Old["Before — trusted to comments"]
        o1["flat map of hand-picked\nmagic sortValues"]
    end
    subgraph NewD["After — derived by construction"]
        n1["ordered family-priority list"]
        n2["sortValue derived off\nSCOPED_FLOOR / SCOPED_CEILING\n+ fixed FAMILY_SPACING"]
        n1 --> n2
        n3["spacing > max combine increment\n⇒ merged lines never cross bands"]
        n2 --> n3
    end
    Old --> NewD
```

### Before vs after — session list order

```mermaid
flowchart LR
    subgraph Before
        direction TB
        b1["Busiest session ever"]
        b2["★ Rare species — Wryneck"]
        b3["First ever Firecrest"]
        b1 --> b2 --> b3
    end
    subgraph AfterList["After"]
        direction TB
        c1["First ever Firecrest"]
        c2["★ Rare species — Wryneck"]
        c3["Busiest session ever"]
        c1 --> c2 --> c3
    end
    Before --> AfterList
```

Keeps the explicit-`sortValue`-per-object design (#369) but replaces the flat magic-number map with two ordered priority bands derived off the scoped range.

wheresrhys:
## Clearer take on the sort mechanism (replacing the confusing section above)

Think of every highlight's `sortValue` as a position on one number line. Scoped records live in a fixed middle range `[SCOPED_FLOOR … SCOPED_CEILING]`. The promoted families need to sit **strictly above** that range and the trailing families **strictly below** it — for *any* possible scoped value, including combined ones.

```mermaid
flowchart TB
    Top["▲ higher sortValue = shown first"]
    L4["only-ever          = CEILING + 4·SPACING"]
    L3["first-ever         = CEILING + 3·SPACING"]
    L2["rare-species       = CEILING + 2·SPACING"]
    L1["long-absence       = CEILING + 1·SPACING"]
    CEIL["── SCOPED_CEILING (6) ──"]
    MID["scoped records\n(real max ≈ 5.13, plus combine bumps)"]
    FLOOR["── SCOPED_FLOOR ──"]
    T1["since-comparison   = FLOOR − 1·SPACING"]
    T2["first-of-year      = FLOOR − 2·SPACING"]
    T3["weight-record      = FLOOR − 3·SPACING"]
    Top --> L4 --> L3 --> L2 --> L1 --> CEIL --> MID --> FLOOR --> T1 --> T2 --> T3
```

Why this is better than before:

```mermaid
flowchart LR
    subgraph Before
        direction TB
        b1["flat map of hand-picked numbers"]
        b2["'above a record' only true if the\nnumbers happen to be right — trusted to comments"]
        b1 --> b2
    end
    subgraph After
        direction TB
        a1["two ordered priority lists\n(leading band / trailing band)"]
        a2["position = CEILING/FLOOR ± index·SPACING"]
        a3["SPACING > max combine bump ⇒\na merged record can never climb into a band"]
        a1 --> a2 --> a3
    end
    Before --> After
```

The guarantee ("always above / always below a record") now holds **by construction** from the band index, instead of depending on someone picking magic numbers that don't overlap.

PR #434, opened 2026-08-09
feat: add top species section to home page

### LINKED ISSUE #427: Add a top birds section to home page
Position: Immediately below recent sessions
Heading: Top species
UI: These should use badge links, linking off to the species pages
Filter: The top 10 species by bird count (potentially a simple way to get this is to use the supabase query method, selecting `id, name, birds (count )` or similar, and sorting by count in js - this avoids creating yet another RPC

wheresrhys:
## Summary
- Adds a "Top species" section to the home page, immediately below Recent Sessions: badge links to the top 10 species by bird count for the viewed group, each linking to its species page.
- Fetches directly in the server-rendered page (`app/(routes)/page.tsx`) via `fetchTopSpecies()` — a `Species` select embedding `birds:Birds(count)`, sorted/limited to top 10 in JS — per the ticket's suggestion to avoid a new RPC.
- Also updates `CLAUDE.md`'s data-fetching convention: fetching is allowed in server actions **or** server-rendered pages/components, not only `app/actions/` — client components remain the one place it's forbidden. Types for a query result scoped to a single page can live alongside that page instead of `app/models/db.ts`.

Closes #427

## Test plan
- [x] `npm run qa` (lint, typecheck, app tests) — passing
- [x] New tests: `fetchTopSpecies` sort/filter/limit behaviour and the rendered "Top species" section, exercised via `app/(routes)/__tests__/page.test.tsx` (heading, badge links sorted/limited/excluding zero-count species, empty state)
- [x] Pre-push hook: full `npm run qa` + DB integration tests + e2e suite all green

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,11 +96,11 @@ The authoritative schema lives in `supabase/schema/` as declarative SQL files, o
 
 ## Data fetching conventions
 
-- All data fetching happens in **server actions** (`app/actions/`), never in client components.
-- Every server action calls `getAuthenticatedSupabaseClient()` to get a group-scoped client.
+- Data fetching happens in **server actions** (`app/actions/`) or in server-rendered pages/components — never in client components (anything marked `'use client'`).
+- Every data-fetching function calls `getAuthenticatedSupabaseClient()` to get a group-scoped client.
 - Errors from Supabase calls are handled via `catchSupabaseErrors()`.
 - RPC calls go through `.rpc('function_name', args)` on the Supabase client.
-- TypeScript types for DB rows come from `app/models/db.ts`, which re-exports from the auto-generated types.
+- TypeScript types for DB rows come from `app/models/db.ts`, which re-exports from the auto-generated types. Types for a query's return shape (e.g. an embedded-relation select used by only one page) can live alongside the page/component that fetches and renders it instead, if not reused elsewhere.
 
 ## Code conventions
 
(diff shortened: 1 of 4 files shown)


wheresrhys:
## Top species selection: filter → sort → slice

`fetchTopSpecies()` queries **all** `Species` rows with an embedded `Birds(count)`, then does the filtering/ordering in JS rather than in the query — so it's worth spelling out what order those three steps run in, since getting it wrong (e.g. slicing before filtering zero-count species) would silently drop a caught species from the badge list.

```mermaid
flowchart TB
    A["Species select + Birds(count)\ne.g. 13 species, incl. Jay: 0 birds"] --> B["filter: count > 0\nJay dropped -> 12 species"]
    B --> C["sort: count descending\nBlackbird 42 ... Coal Tit 1"]
    C --> D["slice: top 10\nGreenfinch (2), Coal Tit (1) dropped"]
    D --> E["10 badge links rendered\nBlackbird ... Nuthatch"]
```

| Species | Bird count | In top 10? |
|---|---|---|
| Blackbird | 42 | yes (1st) |
| ... | ... | yes |
| Nuthatch | 4 | yes (10th) |
| Greenfinch | 2 | no — 11th |
| Coal Tit | 1 | no — 12th |
| Jay | 0 | no — filtered out before ranking |

Zero-count species are excluded *before* ranking, so a species with 0 birds never occupies a top-10 slot even transiently — it can't push out a real 10th-place species.

PR #435, opened 2026-08-09
test: isolate DB integration test writes from concurrent worktree runs

### LINKED ISSUE #419: Audit supabase/__tests__/ write tests for identifier collisions; isolate with random/ticket-specific values
## Context

`swarm` now runs tickets concurrently, each in its own git worktree, against the **same shared
local Supabase instance** (there's only one). `implement-ticket`'s new rule requires any
new/edited DB integration test (`supabase/__tests__/`) to use random or ticket/branch-specific
identifiers for rows it writes, so concurrent runs never collide. Existing tests predate that
rule and haven't been audited.

Confirmed collision-risk literals already in the write paths:
- `readonly-role.test.ts` inserts a species named `'Readonly Test Species'`, renames a group to
  `'Readonly Test Rename'`.
- `triggers.test.ts` already partially isolates via a `TRIG-TEST-*` ring-number prefix — worth
  extending as the convention rather than inventing a new one.
- `rpc-functions.test.ts` has several `.insert()` blocks with fixed literal values.

- Audit every `.insert()` / `.update()` / `.upsert()` / `.delete()` call in
  `supabase/__tests__/*.ts` for hardcoded literal identifiers used to create or target rows.
- Replace collision-risk literals with random or run-specific identifiers (e.g. a
  `crypto.randomUUID()`-derived suffix, or extending the existing `TRIG-TEST-*`-style prefix
  convention), while keeping tests readable and any values asserted against in expectations
  consistent.
- Confirm cleanup (`afterAll`/`afterEach` deletes) still targets the correct rows after
  identifiers become dynamic.
- `npm run test:integration` passes locally afterwards.
- Briefly document the isolation convention/helper used at the top of any file it's introduced
  in, if not already covered by CLAUDE.md's testing section.

## Out of scope

- Read-only tests against fixed e2e seed data (e.g. reading group `'Alpha'`/`'Beta'`/`'Delta'`
  by name) — not writes, not in scope unless the audit finds one reusing a collision-risk value.
- Changing the `db:seed:e2e` fixtures themselves.
- Building a generic parallel-test-isolation framework — a small per-file helper/convention is
  enough (YAGNI).

## Dependencies

None — can run any time.

## Reuse

Extend `triggers.test.ts`'s existing `TRIG-TEST-*` prefix convention rather than inventing a new
isolation mechanism.

## Stack layers touched

Tests only (`supabase/__tests__/`).

## Commit breakdown

Likely one commit (tests-only, well under 400 LOC) — split per file only if the diff grows past
that.

## Tests

No new test titles — this ticket modifies existing tests for isolation, not new behaviour.
Acceptance is `npm run test:integration` passing after the change, with identifiers no longer
collision-prone under concurrent runs.

wheresrhys:
## Summary

Audits every `.insert()` / `.update()` / `.upsert()` / `.delete()` call in `supabase/__tests__/*.ts` for hardcoded literal identifiers that could collide when `swarm` runs multiple worktrees concurrently against the same shared local Supabase instance.

- Adds a small shared helper module, `supabase/__tests__/test-isolation.ts`, with `randomTestSuffix()`, `randomFutureDate()`, and `addDays()` — documented at the top of the file per CLAUDE.md's DB integration tests section.
- `triggers.test.ts`: extends the existing `TRIG-TEST-*` prefix with a random suffix, and randomises the test location name.
- `rpc-functions.test.ts` (`unweighed encounters` and `juvenile encounters` blocks under `stats_per_day_and_species`): randomises location names, ring numbers, and — importantly — session dates. These blocks feed a group-wide aggregate RPC, so two concurrent runs picking the same fixed date (e.g. `2098-06-01`) would combine their rows into a single aggregate result and break both runs' assertions, even without a unique-constraint violation.
- `readonly-role.test.ts`: randomises the two literal names used in its rejected INSERT/UPDATE, for consistency with the convention (these writes always fail with the read-only transaction error `25006` before committing, so there's no real collision risk today, but the values shouldn't rely on that invariant holding).
- `rls.test.ts` and `ring-sequences.test.ts` contain no writes — out of scope.
- Updates CLAUDE.md's "DB integration tests" section to point at the new shared helper and explain the aggregate-RPC collision case.

## Test plan

- `npm run test:integration` — 5 files, 79 tests, all passing.
- `npm run qa` (lint + type-check + app tests) — 46 files, 561 tests, all passing.
- Pre-push hook (app tests, DB integration tests, and the full Playwright E2E suite) — all green (98 E2E tests passed).

Closes #419

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -194,7 +194,11 @@ Test RPC functions and RLS policies against the real local database. Require `np
 Tests can run concurrently in separate git worktrees (see `swarm`) against the same shared local
 Supabase instance. Any row a write test creates (ring numbers, group names, session/location
 names, etc.) must use a random or ticket/branch-specific identifier — never a fixed literal —
-so parallel runs never collide on the same row.
+so parallel runs never collide on the same row, and so date-based rows never collide in a
+group-wide aggregate RPC's results (e.g. `stats_per_day_and_species`) even without a unique
+constraint. Use the shared helpers in `supabase/__tests__/test-isolation.ts`
+(`randomTestSuffix`, `randomFutureDate`, `addDays`) rather than inventing a new isolation
+mechanism per file; extend an existing prefix convention (e.g. `TRIG-TEST-`) with the suffix.
 
 ## Environment variables
 
--- a/supabase/__tests__/readonly-role.test.ts
+++ b/supabase/__tests__/readonly-role.test.ts
@@ -15,6 +15,7 @@
 import { describe, it, beforeAll, afterAll, expect } from 'vitest';
 import { getAuthenticatedSupabaseClientForGroup } from '../../lib/group-auth';
 import { supabase } from '../../lib/supabase';
+import { randomTestSuffix } from './test-isolation';
 import type { SupabaseClient } from '@supabase/supabase-js';
 
 async function getGroupIdByName(name: string): Promise<number> {
@@ -66,16 +67,20 @@ describe('app_readonly role', () => {
 	});
 
 	it('rejects INSERT with a read-only transaction error', async () => {
+		// The transaction is read-only, so this insert never actually commits — the
+		// random suffix is just for consistency with the isolation convention (see
+		// CLAUDE.md's "DB integration tests" section) in case that guarantee ever changes.
 		const { error } = await readonlyClient
 			.from('Species')
-			.insert({ species_name: 'Readonly Test Species' });
+			.insert({ species_name: `Readonly Test Species ${randomTestSuffix()}` });
 		expect(error?.code).toBe(READ_ONLY_TRANSACTION_ERROR_CODE);
 	});
 
 	it('rejects UPDATE with a read-only transaction error', async () => {
+		// As above: the transaction is read-only, so this update never actually commits.
 		const { error } = await readonlyClient
 			.from('RingingGroups')
-			.update({ group_name: 'Readonly Test Rename' })
+			.update({ group_name: `Readonly Test Rename ${randomTestSuffix()}` })
 			.eq('id', alphaId);
 		expect(error?.code).toBe(READ_ONLY_TRANSACTION_ERROR_CODE);
 	});
--- a/supabase/__tests__/rpc-functions.test.ts
+++ b/supabase/__tests__/rpc-functions.test.ts
@@ -12,6 +12,7 @@ import { describe, it, beforeAll, afterAll, expect } from 'vitest';
 import { execSync } from 'child_process';
 import { getAuthenticatedSupabaseClientForGroup } from '../../lib/group-auth';
 import { supabase } from '../../lib/supabase';
+import { addDays, randomFutureDate, randomTestSuffix } from './test-isolation';
 import type { SupabaseClient } from '@supabase/supabase-js';
 
 async function getGroupIdByName(name: string): Promise<number> {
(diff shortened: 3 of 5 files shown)


wheresrhys:
## Behaviour change: isolating DB integration test writes across concurrent worktrees

The interesting risk here isn't a duplicate-row DB error — it's **silent aggregate contamination**. Two `swarm` worktrees running `rpc-functions.test.ts` at the same time both created Delta-group sessions on the same fixed date (`2098-06-01`), each in their own new `Locations` row. No unique-constraint violation fires (`Sessions` is only unique on `(visit_date, location_id)`), but `stats_per_day_and_species` aggregates **across all of a group's locations for a date**, so both runs' encounters land in one combined row and both runs' counts stop matching.

```mermaid
flowchart TB
    subgraph before["Before — fixed date, both worktrees"]
        direction TB
        A1["Worktree A<br/>Delta · Location A<br/>2098-06-01 · 2 encounters"]
        A2["Worktree B<br/>Delta · Location B<br/>2098-06-01 · 2 encounters"]
        A1 --> AGG["stats_per_day_and_species<br/>groups by (group, date)"]
        A2 --> AGG
        AGG --> BAD["visit_date 2098-06-01<br/>encounter_count = 4"]
        BAD --> FAIL["expect(encounter_count).toBe(2)<br/>FAILS in both worktrees"]
    end
```

```mermaid
flowchart TB
    subgraph after["After — randomFutureDate() per run"]
        direction TB
        B1["Worktree A<br/>Delta · Location A′<br/>2091-07-14 · 2 encounters"]
        B2["Worktree B<br/>Delta · Location B′<br/>2087-02-03 · 2 encounters"]
        B1 --> AGG2["stats_per_day_and_species"]
        B2 --> AGG2
        AGG2 --> OK1["visit_date 2091-07-14<br/>encounter_count = 2 ✓"]
        AGG2 --> OK2["visit_date 2087-02-03<br/>encounter_count = 2 ✓"]
    end
```

Ring numbers (`Birds.ring_no` is `UNIQUE`) and location names (`Locations.location_name` is `UNIQUE`) already had a hard DB constraint forcing a collision to surface as an error — the date case was the quiet one, since it only breaks an aggregate's *value*, not its *existence*. The `visit_date` assertions moved from literal strings (`'2098-06-01'`) to the `visitDates[]` / `visitDate` variable actually written, so the fix and the assertion can never drift apart.

### Isolation approach by file

| File | What was fixed literal before | Now |
|---|---|---|
| `triggers.test.ts` | `TRIG-TEST-N/S/FRESH`, `'Trigger Test Location'` | `TRIG-TEST-*-<suffix>`, `Trigger Test Location <suffix>` |
| `rpc-functions.test.ts` (unweighed/juvenile blocks) | `STATSTEST1..3`, `JUVTEST1..5`, fixed location names, fixed dates | suffixed ring numbers/location names + `randomFutureDate()`/`addDays()` |
| `readonly-role.test.ts` | `'Readonly Test Species'`, `'Readonly Test Rename'` | suffixed — belt-and-braces only, since these inserts/updates always fail with the read-only-transaction error (`25006`) before committing |
| `rls.test.ts`, `ring-sequences.test.ts` | — | no writes, out of scope |

New shared helper: `supabase/__tests__/test-isolation.ts` — `randomTestSuffix()`, `randomFutureDate()`, `addDays()`.

Issue #392, opened 2026-07-18
Stop 'resightings' (records that are not N or S type) coming up as sessions


wheresrhys:
Tracking issue for excluding passive resightings/recoveries from behaving like ringing sessions.

## Design (agreed)

`Encounters.record_type` values other than `N`/`S` (confirmed in real data: `C`, `F`, `T`) represent
passive recoveries/resightings, not a ringing visit. Today every row — regardless of type —
upserts into `Sessions` keyed on `(visit_date, location_id)`, so a single resighting shows up on
the `/sessions` calendar and gets its own session page exactly like a real ringing outing.

Fix: add `Sessions.is_resighting_only boolean NOT NULL DEFAULT FALSE`, and make it part of the
session's **identity**, not a derived/recomputed attribute — the unique constraint becomes
`(visit_date, location_id, is_resighting_only)`. A real ringing catch and an unrelated resighting
on the same day at the same site become two separate `Sessions` rows. Each CSV row picks its own
bucket at import time from its own `record_type` (`NOT IN ('N','S')` → `is_resighting_only = TRUE`)
— no trigger, no cross-row aggregation, no import-order sensitivity. Existing data needs a one-off
migration to split any already-mixed `Sessions` rows.

Consequences of the flag:
- Resighting-only sessions are excluded from **day/session-level** stats (busiest/quietest day,
  first-ever-species-per-day, weight records, `session_count`/`session_effort` in
  `aggregate_stats`), but per-species/per-bird totals (`species_count`, `bird_count`,
  `new_bird_count`, etc.) stay untouched — a resighting still counts as a real encounter of that
  species/bird.
- Resighting-only sessions are hidden from the `/sessions` calendar, don't get their own session
  page (falls through to the existing "no session that day" empty state), and are skipped by
  prev/next session-date navigation.
- A new `/resightings` page lists all non-N/S encounters, optionally grouped by species.
- BTO recovery fields `finding_condition`/`finding_circumstances` (present in the CSV, not stored
  today) are out of scope for the first pass of the resightings page — deferred to a follow-up
  ticket.

Broken into sub-issues; see linked tickets for the implementation sequence and blocked-by graph.

## Background (original notes)

Needs some thought as to how to tweak the model/import etc, but eventually would like at least this

- the data to be used as any other records in analysis & stats
- a page which lists all of these non-S recoveries, optionally grouped by species
- they DO NOT appear as sessions having their own session page
- Maybe the model should change so that
[...]

PR #436, opened 2026-08-09
Add Sessions.is_resighting_only identity column and split mixed sessions on import

### LINKED ISSUE #428: Add Sessions.is_resighting_only identity column and split mixed sessions on import
## Context
Part of #392: `Encounters.record_type` values other than `N`/`S` (real data has `C`, `F`, `T`) represent passive recoveries/resightings, not a ringing visit, but today every row upserts into `Sessions` keyed only on `(visit_date, location_id)` (`supabase/schema/schemas/public/tables/"Sessions".sql`, constraint `Sessions_visit_date_location_id_key`), so a resighting shares a `Sessions` row with any real ringing catch on the same day/site and shows up on `/sessions` exactly like a real outing.

This ticket lays the foundation the other 4 sub-tickets of #392 build on: it makes `is_resighting_only` part of a session's *identity* (not a derived/recomputed flag). Each `Encounters` row's own `record_type` deterministically picks its session bucket at import time — no trigger, no cross-row aggregation, no import-order sensitivity.

- `supabase/schema/schemas/public/tables/"Sessions".sql`:
  - Add column `is_resighting_only boolean NOT NULL DEFAULT FALSE` to the `CREATE TABLE public."Sessions"` block.
  - Drop `ALTER TABLE public."Sessions" ADD CONSTRAINT "Sessions_visit_date_location_id_key" UNIQUE (visit_date, location_id);` and replace with `UNIQUE (visit_date, location_id, is_resighting_only)`.
  - No change to `trg_set_session_generated_fields` — it only derives `ringing_group_id` from `location_id` and fires on every INSERT regardless of client, so it still runs correctly for rows inserted by the data migration below.
- `lib/demon-import.ts`, `processEncounterRow` (currently lines 182–245):
  - Compute `const isResightingOnly = !['N', 'S'].includes(row.record_type as string);` before the `Sessions` upsert.
  - Change the `Sessions` upsert call's data object to `{ visit_date: visitDate, location_id: locationId, is_resighting_only: isResightingOnly }` and its conflict target from `'visit_date,location_id'` to `'visit_date,location_id,is_resighting_only'`.
- One-off hand-written data migration, appended as raw SQL to the bottom of the generated migration file (moves existing rows, not just DDL):
  1. For every existing `Sessions` row whose linked `Encounters` mix `record_type IN ('N','S')` with `record_type NOT IN ('N','S')`: insert one new `Sessions` row per such original row (same `visit_date`/`location_id`, `is_resighting_only = TRUE`, letting `trg_set_session_generated_fields` derive `ringing_group_id`), then repoint that row's non-N/S `Encounters.session_id` to the new row.
  2. For any remaining `Sessions` rows that are now (or already were) purely resighting-only: `UPDATE "Sessions" SET is_resighting_only = TRUE` in place.
  3. Sessions with zero `Encounters`, or whose `Encounters` are all `N`/`S`, are untouched.
- Regenerate `types/supabase.types.ts` via `npm run db:types`.
- `lib/__tests__/demon-import.test.ts`: update assertions that hard-code the old conflict target/data shape.

## Out of scope
- Any change to `aggregate_stats`, `stats_per_day_and_species`, session-page/calendar filtering, or the new `/resightings` page — separate sub-tickets of #392.
- `finding_condition`/`finding_circumstances` BTO fields — deferred per #392.
- Any UI change.

## Dependencies
None — base ticket for the #392 breakdown.

## Reuse
- `createUpserter` (`lib/demon-import.ts`) already takes a generic `uniqueColumn`/conflict-target string — widening it to `'visit_date,location_id,is_resighting_only'` is a pure call-site change.
- Follow the declarative-schema workflow in this repo's `CLAUDE.md`: edit the `Sessions.sql` schema file, `npm run db:schema:apply` to generate the migration, hand-append the data-migration SQL to the generated file, `npm run db:seed:local`, then `npm run db:types`.
- `supabase/__tests__/triggers.test.ts` / `ring-sequences.test.ts` show this repo's integration-test convention (branch/ticket-scoped literal identifiers to avoid parallel-worktree collisions).

## Stack layers touched
Database schema + hand-written data migration, shared import logic (`lib/demon-import.ts`), generated types, unit + DB integration tests. No server actions/components/pages.

1. Schema + migration: add column/constraint, generate migration, hand-append data-migration SQL, seed, regenerate types.
2. Import wiring: compute `isResightingOnly` in `processEncounterRow`, update the `Sessions` upsert call.
3. Unit test updates for the new conflict target/data shape + one test per `record_type` bucket.
4. DB integration test exercising the unique constraint and the split/flip migration behaviour.

## Tests (USE algorithm)
`lib/__tests__/demon-import.test.ts` — `describe('processEncounterRow')`:
- Usual: upserts Session with `is_resighting_only: false` for an `N` record_type (update existing test); same for `S`.
- Structure: upserts Session with `is_resighting_only: true` for `C`, `F`, and `T` record_types (one test each).
- Edge: treats an unrecognised/future record_type as resighting-only (guards the inversion rather than an allow-list).

New `supabase/__tests__/` integration test — `describe('Sessions — is_resighting_only identity')`:
- Usual: allows two Sessions rows with the same visit_date/location_id when is_resighting_only differs; rejects a duplicate row with the same triple.
- Structure: derives ringing_group_id correctly for a hand-inserted resighting-only Sessions row.

Data migration verification — `describe('is_resighting_only backfill migration')`:
- Usual: splits a Sessions row with mixed N/S and non-N/S Encounters into two rows, repointing the non-N/S Encounters.
- Edge: flips an already-pure resighting-only row in place without creating a new row; leaves an all-N/S row at FALSE; leaves a zero-Encounters row at FALSE.

wheresrhys:
## What this covers

Foundation for #392: makes `is_resighting_only` part of a `Sessions` row's identity so a passive resighting/recovery never shares a session with a real ringing catch.

- **Schema** (`supabase/schema/schemas/public/tables/"Sessions".sql`): add `is_resighting_only boolean NOT NULL DEFAULT FALSE`; widen the unique constraint to `(visit_date, location_id, is_resighting_only)`.
- **Import** (`lib/demon-import.ts`): `processEncounterRow` derives `is_resighting_only = !['N','S'].includes(record_type)` from each row's own `record_type` and upserts the Session on the widened conflict target. No trigger, no cross-row aggregation, no import-order sensitivity.
- **Data migration** (hand-appended to the generated migration): splits any pre-existing `Sessions` row whose Encounters mix N/S with non-N/S into two rows (repointing the non-N/S Encounters onto a new resighting-only twin, letting `trg_set_session_generated_fields` derive `ringing_group_id`), then flips any remaining pure-resighting rows in place. Zero-encounter and all-N/S rows are left at FALSE.
- **Types**: regenerated `types/supabase.types.ts`.
- **Tests**: unit tests for the bucketing (N/S -> false; C/F/T + unrecognised -> true) and a new DB integration test (`supabase/__tests__/is-resighting-only.test.ts`) covering the unique constraint, group-id derivation, and the split/flip/leave backfill behaviour.

Closes #428

## Notes / assumptions

- This repo **gitignores `supabase/migrations/`** (commit `799f12c` "stop committing supabase migrations") — the declarative schema is the tracked source of truth. The generated migration carrying the DDL **plus the hand-written data-migration SQL** therefore cannot be committed; it has been placed in the local `supabase/migrations/` directory for the human to inspect and `npm run db:migration:push`. Verify that file is present before deploying.
- `db-migration` label: this ticket touches `supabase/schema/`.
- Out of scope (separate #392 sub-tickets): `aggregate_stats` / `stats_per_day_and_species` changes, `/sessions` calendar & session-page filtering, the `/resightings` page, and `finding_condition`/`finding_circumstances`.

## Test status

- `npm run qa` (lint + type-check + 567 app tests): pass
- `npm run test:integration` (85 tests incl. 6 new): pass
- pre-push hook (app + integration tests): pass

### DIFF
--- a/lib/__tests__/demon-import.test.ts
+++ b/lib/__tests__/demon-import.test.ts
@@ -281,9 +281,74 @@ describe('processEncounterRow', () => {
 		const locationId = 30; // third call returns 30
 		expect(upsert).toHaveBeenCalledWith(
 			'Sessions',
-			{ visit_date: '2023-03-15', location_id: locationId },
-			'visit_date,location_id'
+			{
+				visit_date: '2023-03-15',
+				location_id: locationId,
+				is_resighting_only: false
+			},
+			'visit_date,location_id,is_resighting_only'
 		);
+	});
+
+	describe('is_resighting_only bucketing', () => {
+		function sessionUpsertArgs() {
+			return upsert.mock.calls.find(([table]) => table === 'Sessions');
+		}
+
+		it('buckets an N record_type as is_resighting_only: false', async () => {
+			await processEncounterRow(
+				makeDemonRow({ record_type: 'N' }),
+				upsert,
+				RINGING_GROUP_ID
+			);
+			expect(sessionUpsertArgs()).toEqual([
+				'Sessions',
+				expect.objectContaining({ is_resighting_only: false }),
+				'visit_date,location_id,is_resighting_only'
+			]);
+		});
+
+		it('buckets an S record_type as is_resighting_only: false', async () => {
+			await processEncounterRow(
+				makeDemonRow({ record_type: 'S' }),
+				upsert,
+				RINGING_GROUP_ID
+			);
+			expect(sessionUpsertArgs()).toEqual([
+				'Sessions',
+				expect.objectContaining({ is_resighting_only: false }),
+				'visit_date,location_id,is_resighting_only'
+			]);
+		});
+
+		it.each(['C', 'F', 'T'])(
+			'buckets a %s record_type as is_resighting_only: true',
+			async (recordType) => {
+				await processEncounterRow(
+					makeDemonRow({ record_type: recordType }),
+					upsert,
+					RINGING_GROUP_ID
+				);
+				expect(sessionUpsertArgs()).toEqual([
+					'Sessions',
+					expect.objectContaining({ is_resighting_only: true }),
+					'visit_date,location_id,is_resighting_only'
+				]);
+			}
+		);
+
+		it('treats an unrecognised/future record_type as is_resighting_only: true', async () => {
+			await processEncounterRow(
+				makeDemonRow({ record_type: 'Z' }),
+				upsert,
+				RINGING_GROUP_ID
+			);
+			expect(sessionUpsertArgs()).toEqual([
+				'Sessions',
+				expect.objectContaining({ is_resighting_only: true }),
+				'visit_date,location_id,is_resighting_only'
+			]);
+		});
 	});
 
 	it('upserts Encounter with bird and session IDs from prior upserts', async () => {
--- a/lib/__tests__/demon-import.test.ts
+++ b/lib/__tests__/demon-import.test.ts
@@ -295,33 +295,7 @@ describe('processEncounterRow', () => {
 			return upsert.mock.calls.find(([table]) => table === 'Sessions');
 		}
 
-		it('buckets an N record_type as is_resighting_only: false', async () => {
-			await processEncounterRow(
-				makeDemonRow({ record_type: 'N' }),
-				upsert,
-				RINGING_GROUP_ID
-			);
-			expect(sessionUpsertArgs()).toEqual([
-				'Sessions',
-				expect.objectContaining({ is_resighting_only: false }),
-				'visit_date,location_id,is_resighting_only'
-			]);
-		});
-
-		it('buckets an S record_type as is_resighting_only: false', async () => {
-			await processEncounterRow(
-				makeDemonRow({ record_type: 'S' }),
-				upsert,
-				RINGING_GROUP_ID
-			);
-			expect(sessionUpsertArgs()).toEqual([
-				'Sessions',
-				expect.objectContaining({ is_resighting_only: false }),
-				'visit_date,location_id,is_resighting_only'
-			]);
-		});
-
-		it.each(['C', 'F', 'T'])(
+		it.each(['U', 'F', 'D'])(
 			'buckets a %s record_type as is_resighting_only: true',
 			async (recordType) => {
 				await processEncounterRow(
@@ -337,7 +311,23 @@ describe('processEncounterRow', () => {
 			}
 		);
 
-		it('treats an unrecognised/future record_type as is_resighting_only: true', async () => {
+		it.each(['N', 'S', 'C', 'T'])(
+			'buckets a %s record_type as is_resighting_only: false',
+			async (recordType) => {
+				await processEncounterRow(
+					makeDemonRow({ record_type: recordType }),
+					upsert,
+					RINGING_GROUP_ID
+				);
+				expect(sessionUpsertArgs()).toEqual([
+					'Sessions',
+					expect.objectContaining({ is_resighting_only: false }),
+					'visit_date,location_id,is_resighting_only'
+				]);
+			}
+		);
+
+		it('treats all other record_type as is_resighting_only: false', async () => {
 			await processEncounterRow(
 				makeDemonRow({ record_type: 'Z' }),
 				upsert,
@@ -345,7 +335,7 @@ describe('processEncounterRow', () => {
 			);
 			expect(sessionUpsertArgs()).toEqual([
 				'Sessions',
-				expect.objectContaining({ is_resighting_only: true }),
+				expect.objectContaining({ is_resighting_only: false }),
 				'visit_date,location_id,is_resighting_only'
 			]);
 		});
--- a/lib/demon-import.ts
+++ b/lib/demon-import.ts
@@ -209,10 +209,16 @@ export async function processEncounterRow(
 
 	const visitDate = convertDateFormat(row.visit_date as string);
 
+	const isResightingOnly = !['N', 'S'].includes(row.record_type as string);
+
 	const sessionId = await upsert<SessionsInsert>(
 		'Sessions',
-		{ visit_date: visitDate, location_id: locationId },
-		'visit_date,location_id' as keyof SessionsInsert
+		{
+			visit_date: visitDate,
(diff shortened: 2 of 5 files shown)


wheresrhys:
## Behaviour change: a resighting no longer collapses into a real ringing session

**Before** — `Sessions` was keyed on `(visit_date, location_id)`, so a passive resighting (`record_type` `C`/`F`/`T`) on the same day/site as a real catch (`N`/`S`) shared one Sessions row and appeared on `/sessions` like a real outing.

**After** — the key becomes `(visit_date, location_id, is_resighting_only)`, and each Encounter picks its own bucket at import time from its own `record_type`. Same day/site now yields two distinct rows.

```mermaid
flowchart TB
  subgraph before["BEFORE — key (visit_date, location_id)"]
    B1["Encounter N (real catch)"] --> BS["Session 2024-05-10 @ Reedbed"]
    B2["Encounter C (resighting)"] --> BS
    BS --> BOUT["one session — resighting shows as a real outing"]
  end

  subgraph after["AFTER — key (visit_date, location_id, is_resighting_only)"]
    A1["Encounter N (real catch)"] --> AS1["Session 2024-05-10 @ Reedbed<br/>is_resighting_only = false"]
    A2["Encounter C (resighting)"] --> AS2["Session 2024-05-10 @ Reedbed<br/>is_resighting_only = true"]
  end
```

`is_resighting_only` is derived per row as `!['N','S'].includes(record_type)` — so any unrecognised/future code buckets as a resighting (guards the inversion rather than relying on an allow-list of `C`/`F`/`T`).

## Backfill: one-off migration splits already-mixed sessions

For existing data, a Sessions row whose Encounters mix N/S with non-N/S is split: a new resighting-only twin is inserted (same day/site) and the non-N/S Encounters are repointed onto it. Pure-resighting rows are flipped in place; all-N/S and zero-encounter rows are untouched.

```mermaid
flowchart TB
  M["Session #7 @ 2024-05-10<br/>is_resighting_only = false<br/>Encounters: N (bird A), C (bird B)"]
  M --> S7["Session #7 (kept)<br/>is_resighting_only = false<br/>Encounters: N (bird A)"]
  M --> S9["Session #9 (new twin)<br/>is_resighting_only = true<br/>Encounters: C (bird B) — repointed"]
```

| Original session shape | Migration action | Result |
|---|---|---|
| Mixed N/S + non-N/S | split into two rows, repoint non-N/S | original stays FALSE, new twin TRUE |
| All non-N/S | flip in place | TRUE, no new row |
| All N/S | untouched | FALSE |
| Zero encounters | untouched | FALSE |

lib/__tests__/demon-import.test.ts
```diff
@@ -281,11 +281,76 @@ describe('processEncounterRow', () => {
...
+				RINGING_GROUP_ID
+			);
+			expect(sessionUpsertArgs()).toEqual([
+				'Sessions',
+				expect.objectContaining({ is_resighting_only: false }),
+				'visit_date,location_id,is_resighting_only'
+			]);
+		});
+
+		it.each(['C', 'F', 'T'])(
```
wheresrhys:
```suggestion
		it.each(['U', 'F', 'D'])(
```

lib/__tests__/demon-import.test.ts
```diff
@@ -281,11 +281,76 @@ describe('processEncounterRow', () => {
...
+				);
+				expect(sessionUpsertArgs()).toEqual([
+					'Sessions',
+					expect.objectContaining({ is_resighting_only: true }),
+					'visit_date,location_id,is_resighting_only'
+				]);
+			}
+		);
+
+		it('treats an unrecognised/future record_type as is_resighting_only: true', async () => {
```
wheresrhys:
```suggestion
		it('treats all other record_type as is_resighting_only: false', async () => {
```

And adjust the test body and the src to implement this behaviour

lib/demon-import.ts
```diff
@@ -209,10 +209,16 @@ export async function processEncounterRow(
 
 	const visitDate = convertDateFormat(row.visit_date as string);
 
+	const isResightingOnly = !['N', 'S'].includes(row.record_type as string);
```
wheresrhys: As indicated here https://github.com/wheresrhys/totf/pull/436/changes#r3744978653, the rule is that U, F, D are resighting only. anything else should be considered a standard session encounter

wheresrhys:
Fixed — thanks for catching this. The classification rule was inverted: `processEncounterRow` was treating everything except `N`/`S` as resighting-only, when the actual rule is that only `record_type` `U`, `F`, `D` are resighting-only and everything else is a standard session encounter.

Changes in 4fc7524:
- `lib/demon-import.ts`: `isResightingOnly` now computed as `['U', 'F', 'D'].includes(record_type)`.
- `lib/__tests__/demon-import.test.ts`: applied both suggested edits — `it.each(['U', 'F', 'D'])(...)` for the true case, and a single `'treats all other record_type as is_resighting_only: false'` test (using `it.each(['N', 'S', 'C', 'T'])` plus an explicit unrecognised-code case) covering the false branch.
- The one-off backfill migration and its integration test (`supabase/__tests__/is-resighting-only.test.ts`) also encoded the old (inverted) rule, so I aligned those too — otherwise the backfill would have split/flipped sessions using the opposite logic from the live importer.

Test status: `npm run test:nowatch` — 569/569 passing. `npm run test:integration` — all `is_resighting_only` tests pass (6/6); 2 pre-existing/unrelated failures in `rpc-functions.test.ts` (`notable_retraps` ARRETRAP `proven_age` assertions) come from a known race in `trg_encounters_refresh_bird_proven_age` during the seed script's 30-concurrent CSV import, not from this change — ARRETRAP's `record_type` values (`N`/`S`) bucket identically under both the old and corrected rule.

PR #437, opened 2026-08-09
Split E2E suite so most branches skip shared-state-mutating specs

wheresrhys:
## Summary
- Concurrent `swarm` worktrees push at the same time and both run the full Playwright E2E suite against the single shared local Supabase instance. `e2e/authenticated/import.spec.ts` does raw, unscoped writes/deletes against fixed seed-data rows, so unrelated specs in a sibling worktree can fail mid-push.
- Tag any spec that mutates shared fixture rows with `@mutates` (currently only `import.spec.ts`) and add `e2e/mutating-spec-triggers.json` mapping that tag to the source paths that make it relevant.
- `.husky/pre-push` now runs `scripts/e2e-select-suite.sh`: diffs the branch against `origin/main`, runs the full `test:e2e` only if the diff touches a trigger path, otherwise `test:e2e:safe` (`--grep-invert @mutates`). Most branches never execute the destructive spec at all — no reseeding, no coordination needed.
- `test:e2e` (full) is unchanged and still what CI uses.
- Fixed a doc/hook mismatch found along the way: CLAUDE.md claimed the pre-push hook ran DB integration tests — it never did; it actually runs E2E tests (previously undocumented as a suite at all). Corrected the suite table and added an "E2E tests (Playwright)" section.
- Also fixed two related contention bugs surfaced while verifying this: `.claude/worktrees/` (where `swarm` puts per-ticket worktrees) was neither gitignored nor excluded from eslint/vitest, so a worktree with build/test artifacts on disk broke lint and app-test runs from the main checkout.

This is increment 1 of 2 — increment 2 will generalize `swarm`'s `db-migration` in-flight cap into a small exclusive-resource label set (`db-migration` + a new `e2e-exclusive`) so the rare ticket that *does* need the mutating spec is still serialized across worktrees, without a filesystem lock.

## Test plan
- [x] `npm run test:e2e:safe` — 95 tests, `import flow` specs excluded
- [x] `npm run test:e2e:mutates` — only the 3 `import flow` tests run, all pass
- [x] `scripts/e2e-select-suite.sh` picks `test:e2e:safe` on a diff untouched by trigger paths, and full `test:e2e` when the diff touches `import.spec.ts`/`lib/demon-import.ts`
- [x] Full pre-push hook run (lint, type-check, `test:nowatch`, `e2e-select-suite.sh`) — 561 app tests + 98 E2E tests passing

### DIFF
--- a/e2e/mutating-spec-triggers.json
+++ b/e2e/mutating-spec-triggers.json
@@ -1,0 +1,8 @@
+{
+	"@mutates": [
+		"lib/demon-import.ts",
+		"app/api/import/",
+		"scripts/import-csv.ts",
+		"e2e/authenticated/import.spec.ts"
+	]
+}
--- a/scripts/resolve-mutating-import-deps.mjs
+++ b/scripts/resolve-mutating-import-deps.mjs
@@ -11,13 +11,17 @@
  * Type-only imports are excluded, since they can't affect runtime behaviour:
  *   - whole-statement `import type { X } from '...'`
  *   - named specifiers marked inline, `import { type X } from '...'`
- *   - named specifiers that resolve to a `export type`/`export interface`
- *     declaration in the target file, even when imported without the `type`
- *     keyword (common in this codebase, e.g. `types/supabase.types.ts`)
+ *   - named specifiers that resolve to a `export type`/`export interface`/
+ *     `export type { X }`/`export { type X }` declaration in the target
+ *     file, even when imported without the `type` keyword (common in this
+ *     codebase, e.g. `types/supabase.types.ts`)
  *
- * This is a regex-based scan, not a full TS parse — it's deliberately kept
- * lightweight (one level deep only) rather than pulling in the TypeScript
- * compiler API for a check this narrow in scope.
+ * Each trigger file (and each target file it imports from) is parsed into a
+ * real TypeScript AST via the `typescript` compiler API, rather than
+ * pattern-matching import statements with regexes — so the parsing handles
+ * the full grammar (multi-line clauses, aliases, JSX, etc.) correctly rather
+ * than approximately. Resolution is still deliberately kept to one level
+ * deep only, not a full transitive dependency graph.
  *
  * Usage: node scripts/resolve-mutating-import-deps.mjs
  * Prints one repo-relative path per line.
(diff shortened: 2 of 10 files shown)


e2e/mutating-spec-triggers.json
```diff
@@ -0,0 +1,8 @@
+{
+	"@mutates": [
+		"lib/demon-import.ts",
```
wheresrhys, line 3: I think this should also include direct src dependencies of these files (excluding things only used to import types)

This might be better done in scripts/e2e-select-suite.sh by resolving the dependency tree in there somehow

wheresrhys:
Addressed in d918d05: added `scripts/resolve-mutating-import-deps.mjs`, which resolves each `@mutates` trigger file's direct (one-level, not transitive) local imports via a lightweight regex scan — excluding type-only imports (`import type ...`, and named imports that resolve to a `export type`/`export interface` in the target file, e.g. `types/supabase.types.ts`'s `Database`). `scripts/e2e-select-suite.sh` now also matches the diff against those resolved dependency paths, so e.g. editing `lib/group-auth.ts` (imported by `app/api/import/route.ts`) or `e2e/helpers/group-ids.ts` (imported by `import.spec.ts`) correctly triggers the full E2E suite even though neither file is itself listed in `mutating-spec-triggers.json`.

Kept it one level deep and regex-based rather than pulling in the TS compiler API, per the "not over-engineered" guidance — verified manually (a file only reachable via a trigger's import now selects the full suite; an unrelated file still selects the safe subset; the type-only exclusion correctly leaves `types/supabase.types.ts` out) and confirmed `npx playwright test --grep @mutates` still passes.

scripts/resolve-mutating-import-deps.mjs
```diff
@@ -0,0 +1,162 @@
...
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const repoRoot = path.resolve(__dirname, '..');
+
+const triggersPath = path.join(repoRoot, 'e2e', 'mutating-spec-triggers.json');
+const triggers = JSON.parse(readFileSync(triggersPath, 'utf8'))['@mutates'];
+
+const IMPORT_STATEMENT_RE =
```
wheresrhys: Please can you use a more robust approach that generates and traverses an AST

wheresrhys:
Addressed in 895987e: `scripts/resolve-mutating-import-deps.mjs` now parses each `@mutates` trigger file (and the files it imports from) with the `typescript` compiler API instead of pattern-matching import statements with regexes.

- Trigger files and their import targets are parsed into a real AST via `ts.createSourceFile`.
- Import declarations are walked as AST nodes (`ImportDeclaration` → `importClause` → `namedBindings`), so multi-line clauses, aliasing, etc. are handled by the actual grammar rather than approximated.
- Type-only detection now inspects the target file's AST for `export type X` / `export interface X` declarations *and* `export type { X }` / `export { type X }` named re-exports — the latter is a case the old regex heuristic missed entirely (it only matched `export type`/`export interface` declaration syntax, not export-specifier lists). Confirmed via a smoke test that `export { type Bar as AliasedBar }` is now correctly excluded.
- Scope is unchanged: still one-level/direct resolution only, not a full transitive dependency graph — that was an intentional design decision, not something this feedback asked to revisit.
- `scripts/e2e-select-suite.sh`'s contract (invocation + one-repo-relative-path-per-line stdout) is unchanged, so it needed no edits. Updated the `CLAUDE.md` E2E section to describe the AST-based mechanism instead of "regex-based import scan".

Verification: ran the new script and confirmed byte-for-byte identical output to the old regex-based version against the current trigger set (`app/actions/group-cookie.ts`, `lib/group-auth.ts`, `lib/demon-import.ts`, `lib/supabase.ts`, `e2e/helpers/group-ids.ts`). Also reproduced `scripts/e2e-select-suite.sh`'s matching logic against three scenarios: a direct-dependency change (selects full suite), an unrelated change (selects safe suite), and a change to a type-only-imported file (`types/supabase.types.ts`, correctly still selects safe suite despite being physically imported by two trigger-adjacent files). `npm run lint`, `tsc --noEmit`, and `npm run test:nowatch` all pass.

PR #450, opened 2026-08-10
Add condition/circumstances filters to the /resightings page

### LINKED ISSUE #433: Add condition/circumstances filters to the /resightings page (raw codes, no decoding)
## Context
Sub-ticket of #392. #432 (blocked-by dependency) adds `finding_condition`/`finding_circumstances` columns to `Encounters` and surfaces them as plain columns on the `/resightings` page (added by #431). Neither field is covered by the DemOn field spec (`demon-ringing-data-entry-fields.xls`) — see updated `CLAUDE.md` note — so their values are opaque codes with no known human-readable mapping; this ticket uses the raw codes as-is (no decoding) to let users narrow the resightings list, the same way the page already lets them narrow by species.

- On `/resightings` (component `app/components/ResightingsTable.tsx`, added by #431/#432), add two filter controls — "Condition" and "Circumstances" — each a `<select>` populated with the distinct raw `finding_condition` / `finding_circumstances` values present in the currently loaded rows, plus an "All" option (default).
- Filters combine with the existing species tab (`AND` semantics): the visible rows are those matching the active species tab AND the selected condition AND the selected circumstances.
- A row with a `null` value for either field is only shown when that filter is set to "All" (it never matches a specific selected code).
- No decoding/lookup table — display and filter on the raw stored string exactly as saved (e.g. `'08'`, `'28'`, `'--'`), per `CLAUDE.md`'s note that these two fields are not covered by the DemOn spec.

## Out of scope
- Any decode/lookup table mapping codes to human-readable labels — explicitly rejected; codes are opaque.
- Filtering/grouping by any field other than `finding_condition`/`finding_circumstances`.
- Server-side filtering/pagination — the resightings page already loads its full result set client-side (per #431); these filters operate on that same in-memory list, consistent with the existing species-tab filtering.

## Dependencies
Blocked by #432 (adds the two columns and selects them in the resightings query).

## Reuse
- `TabNav` / `groupByDiscrepancyType`-style grouping already used for the species tabs (#431) — the new filters are a second, independent narrowing step over the same row list, not a replacement for the species tabs.
- Existing `<select>`-based filter patterns elsewhere in `app/components/` for the control markup/styling (grep for existing filter `<select>` usage before inventing new markup).

## Stack layers touched
Component only (`app/components/ResightingsTable.tsx`) + its tests. No server actions, schema, or page-routing changes.

1. Add the two `<select>` filters (distinct-value derivation + `null`-aware "All" default) and combine with the existing species-tab filtering in `ResightingsTable.tsx`.
2. Tests.

## Tests (USE algorithm)
`app/components/__tests__/ResightingsTable.test.ts` (extend):
- Usual: filtering by a specific condition code shows only rows with that exact `finding_condition` value; same for circumstances.
- Structure: the condition filter's options are exactly the distinct `finding_condition` values present in the data, plus "All"; same for circumstances.
- Structure: condition and circumstances filters combine with each other and with the active species tab (AND semantics) — a row must match all three to show.
- Edge: a row with `finding_condition: null` is hidden once any specific condition code is selected, and reappears when the filter is reset to "All"; same for circumstances.
- Edge: filters render "All" only (no other options) and show every row when no rows have a `finding_condition`/`finding_circumstances` value at all.

## Model label: sonnet
Two client-side `<select>` filters combined with AND logic over an already-loaded row list — no new data-fetching, no schema change, no decoding logic, mirrors an existing tab/filter mechanism already in the codebase. Matches CLAUDE.md's "small, precisely specified, low-risk changes" bar.

wheresrhys:
## Summary

Adds two client-side `<select>` filters — Condition and Circumstances — to `ResightingsTable` (`app/components/ResightingsTable.tsx`), populated with the distinct raw `finding_condition` / `finding_circumstances` values present in the currently loaded rows, plus an "All" option (default). No decode/lookup table — codes are displayed and filtered on as-is, per `CLAUDE.md`'s note that these two fields aren't covered by the DemOn spec.

- Filters combine with the existing species tab using AND semantics.
- A row with a `null` value for either field only shows when that filter is set to "All".
- Reused the existing client-side `<select>` filter pattern from `BirdSearchResults.tsx` (sentinel "All" value + `.filter()` over the already-loaded row list) rather than inventing new markup.

## Assumptions

- The filter select's option lists are derived from the full set of loaded `resightings` (not scoped to the active species tab), so switching species tabs doesn't change what condition/circumstances codes are offered — matches the ticket's "currently loaded rows" wording.
- Both filter selects always render (rather than being hidden when they'd only offer "All"), since the ticket's edge-case test explicitly expects an "All"-only select to render.

Closes #433


wheresrhys:
## Behaviour change: condition/circumstances filters on `/resightings`

Before this PR, `/resightings` only narrowed rows by the species tab. Now two more filters — Condition and Circumstances — combine with the tab using **AND**, so all three must match for a row to show.

```mermaid
flowchart TB
    subgraph before["Before: species tab only"]
        B1["Species tab = Blue Tit"] --> B2["Shows every Blue Tit row,\nregardless of finding_condition\nor finding_circumstances"]
    end

    subgraph after["After: tab AND condition AND circumstances"]
        A1["Species tab = Blue Tit"] --> A4{"Row matches\nALL THREE?"}
        A2["Condition = 08"] --> A4
        A3["Circumstances = 10"] --> A4
        A4 -->|yes| A5["Row shown"]
        A4 -->|no| A6["Row hidden"]
    end
```

### The `null` edge case

The subtle part: a row with `finding_condition: null` (never coded) must **never** match a specific selected code — only "All". Naively treating `null` as a wildcard would make it show up under every code, which is wrong (it means "not recorded", not "matches everything").

```mermaid
flowchart TB
    R["Row: RING3, finding_condition = null"]
    R --> Q1{"Condition filter = All?"}
    Q1 -->|yes| S1["shown"]
    Q1 -->|no, e.g. 08| S2["hidden — null never\nmatches a specific code"]
```

| Condition filter | RING1 (`finding_condition: '08'`) | RING3 (`finding_condition: null`) |
|---|---|---|
| All (default) | shown | shown |
| `08` | shown | **hidden** |

PR #466, opened 2026-08-13
Add DB integration tests for demon-import upsert/uniqueness behaviour

### LINKED ISSUE #464: Add integration tests for demon-import.ts
So much of the success of this module depends on the db schema, so I'd like tests that check that multiple records can be written, and that effectively checks that the uniquness constraints cause errors/allow writes appropriately

wheresrhys:
## What this covers

`lib/demon-import.ts`'s `createUpserter` + `processEncounterRow` lean entirely on each
table's real unique constraints to decide whether an upsert reuses an existing row or
creates a new one — but the existing unit tests (`lib/__tests__/demon-import.test.ts`)
mock the Supabase client entirely, so none of that DB-level behaviour was actually
exercised against the schema.

Adds `supabase/__tests__/demon-import.test.ts`, a DB integration test suite (requires
local Supabase, run via `npm run test:integration`) covering:

- **Usual**: multiple distinct CSV rows write independent Species/Birds/Locations/
  Sessions/Encounters rows; re-importing the same rows a second time is idempotent
  (no duplicates).
- **Structure** (one describe block per table's `uniqueColumns`, matching
  `processEncounterRow`'s upsert calls):
  - `Species` (`species_name`): shared species name resolves to the same row.
  - `Birds` (`ring_no`): same ring number with a different species updates
    `species_id` in place rather than creating a second Bird.
  - `Locations` (`location_name`, `ringing_group_id`): same location name under two
    different groups creates two separate rows (regression coverage for the composite
    key fixed in #465/9cc4602); same name under one group reuses the row.
  - `Sessions` (`visit_date`, `location_id`, `session_type`): differing
    `session_type` creates separate sessions for the same date/location; matching
    triples reuse the same row.
  - `Encounters` (`bird_id`, `session_id`): reprocessing the same bird+session
    updates the existing row in place (exact-duplicate edge case); a different
    session for the same bird creates a new row (near-duplicate edge case).
- Each describe block also asserts a **raw duplicate insert** (bypassing `upsert`,
  going straight through `.insert()`) is rejected with Postgres error `23505` — this
  is the literal "uniqueness constraint causes an error" check from the ticket, and
  catches drift if a table's unique constraint is ever weakened or dropped without
  `demon-import.ts`'s `uniqueColumns` being updated to match.

All rows use a per-run random suffix (`randomTestSuffix`) and random future dates
(`randomFutureDate`/`addDays`) per `supabase/__tests__/test-isolation.ts`, so concurrent
worktree runs against the shared local Supabase instance never collide. `afterAll`
cleans up every row this file could have created.

15 tests, all passing locally (`npm run test:integration`), and the full local
integration suite (126 tests across 7 files) still passes.

## Also includes: unrelated drive-by test fix

`app/(routes)/mistakes/__tests__/page.test.tsx`'s "renders ring_no as a link in each
row" test was failing on `origin/main` HEAD independently of this change (confirmed via
CI on 4d5704f, and by running the test against the unmodified main working copy) — it
compared the first *rendered* table row (sorted by species ascending, per
`MistakesDiscrepancyTable`'s `initialSortColumn`) against the first row in the
snapshot fixture's raw, unsorted array order. Since this failure blocks the pre-push
hook for every branch cut from the current `main`, it's fixed here in a separate commit
(no product code touched) so this PR's own push could go through. No connection to the
demon-import work otherwise.

## Assumptions

- Interpreted the ticket's "checks that the uniqueness constraints cause errors/allow
  writes appropriately" as needing both: (a) upsert-level reuse-vs-create behaviour per
  table, and (b) a literal duplicate-insert-rejected check per table's constraint.
- No `db-migration`/`e2e-exclusive` label needed — this PR touches
  `supabase/__tests__/` only, not `supabase/schema/`, and doesn't touch any path in
  `e2e/mutating-spec-triggers.json`.

Closes #464

### DIFF
--- a/supabase/__tests__/demon-import.test.ts
+++ b/supabase/__tests__/demon-import.test.ts
@@ -314,81 +314,6 @@ describe('demon-import — Species uniqueness (species_name)', () => {
 		);
 		expect(Number(speciesIds)).toBe(1);
 	});
-
-	it('rejects a raw duplicate species_name insert (bypassing upsert) with a unique-violation error', async () => {
-		const speciesName = `DemonImportSpecies-${suffix}-raw-dup`;
-		const first = await groupClient
-			.from('Species')
-			.insert({ species_name: speciesName });
-		expect(first.error).toBeNull();
-
-		const duplicate = await groupClient
-			.from('Species')
-			.insert({ species_name: speciesName });
-		expect(duplicate.error?.code).toBe('23505');
-	});
-});
-
-describe('demon-import — Birds uniqueness (ring_no)', () => {
-	let groupId: number;
-	let groupClient: SupabaseClient;
-
-	beforeAll(async () => {
-		groupId = createIsolatedGroup(`demon-import-${suffix}-birds`);
-		groupClient = await getAuthenticatedSupabaseClientForGroup(groupId);
-	});
-
-	it('upserting two rows with the same ring_no but a different species reuses the same Birds row and updates species_id', async () => {
-		const upsert = createUpserter(groupClient);
-		const ringNo = `DEMON-TEST-${suffix}-bird-shared`;
-
-		await processEncounterRow(
-			makeRow({
-				ring_no: ringNo,
-				species_name: `DemonImportSpecies-${suffix}-bird-1`,
-				loc_id: `DemonImportLoc-${suffix}-bird`
-			}),
-			upsert,
-			groupId
-		);
-		await processEncounterRow(
-			makeRow({
-				ring_no: ringNo,
-				species_name: `DemonImportSpecies-${suffix}-bird-2`,
-				loc_id: `DemonImportLoc-${suffix}-bird`
-			}),
-			upsert,
-			groupId
-		);
-
-		expect(
-			psqlCount(`SELECT COUNT(*) FROM "Birds" WHERE ring_no = '${ringNo}';`)
-		).toBe(1);
-
-		const currentSpeciesName = psqlScalar(
-			`SELECT s.species_name FROM "Birds" b JOIN "Species" s ON s.id = b.species_id WHERE b.ring_no = '${ringNo}';`
-		);
-		expect(currentSpeciesName).toBe(`DemonImportSpecies-${suffix}-bird-2`);
-	});
-
-	it('rejects a raw duplicate ring_no insert (bypassing upsert) with a unique-violation error', async () => {
-		const ringNo = `DEMON-TEST-${suffix}-bird-raw-dup`;
-		const speciesId = Number(
-			psqlScalar(
-				`INSERT INTO "Species" (species_name) VALUES ('DemonImportSpecies-${suffix}-bird-raw-dup') RETURNING id;`
-			)
-		);
-
-		const first = await groupClient
-			.from('Birds')
-			.insert({ ring_no: ringNo, species_id: speciesId });
-		expect(first.error).toBeNull();
-
-		const duplicate = await groupClient
-			.from('Birds')
-			.insert({ ring_no: ringNo, species_id: speciesId });
-		expect(duplicate.error?.code).toBe('23505');
-	});
 });
 
 describe('demon-import — Locations uniqueness (location_name, ringing_group_id)', () => {
@@ -468,19 +393,6 @@ describe('demon-import — Locations uniqueness (location_name, ringing_group_id
 				`SELECT COUNT(*) FROM "Locations" WHERE location_name = '${locationName}';`
 			)
 		).toBe(1);
-	});
-
-	it('rejects a raw duplicate (location_name, ringing_group_id) insert (bypassing upsert) with a unique-violation error', async () => {
-		const locationName = `DemonImportLoc-${suffix}-raw-dup`;
-		const first = await groupClientA
-			.from('Locations')
-			.insert({ location_name: locationName, ringing_group_id: groupIdA });
-		expect(first.error).toBeNull();
-
-		const duplicate = await groupClientA
-			.from('Locations')
-			.insert({ location_name: locationName, ringing_group_id: groupIdA });
-		expect(duplicate.error?.code).toBe('23505');
 	});
 });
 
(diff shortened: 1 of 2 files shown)


wheresrhys:
## What this PR actually verifies

`lib/demon-import.ts` never enforces uniqueness itself — every upsert just tells Postgres
"resolve conflicts on these columns" and trusts the table's real unique constraint to back
that up. This PR adds DB integration tests that prove the trust is warranted, table by table.
Two of those checks are worth visualising because they'd previously fail *silently* (extra
rows, not errors) if the schema and the code's `uniqueColumns` ever drifted apart.

### 1. Locations: two groups importing the same site name

Before the composite `(location_name, ringing_group_id)` constraint (fixed in #465, just
merged to `main`), a single global unique constraint on `location_name` meant two ringing
groups both calling a site "Garden Trap" would collide on the *same* row — group B's import
would silently overwrite group A's location ownership.

```mermaid
flowchart TB
    subgraph old["Old: unique(location_name) only"]
        A1["Group Alpha imports\nloc_id = 'Garden Trap'"] --> R1["Locations row #1\nGarden Trap"]
        A2["Group Beta imports\nloc_id = 'Garden Trap'"] --> R1
        R1 -.->|"ringing_group_id\noverwritten to Beta"| BUG["Alpha's sessions now\npoint at a Beta-owned location"]
    end

    subgraph new["New: unique(location_name, ringing_group_id) — this PR's test"]
        B1["Group Alpha imports\nloc_id = 'Garden Trap'"] --> R2["Locations row #1\nGarden Trap · Alpha"]
        B2["Group Beta imports\nloc_id = 'Garden Trap'"] --> R3["Locations row #2\nGarden Trap · Beta"]
    end
```

The new `Locations uniqueness` describe block asserts this directly: same `location_name`
under two different `ringingClient`s produces **two** rows, not one.

### 2. Encounters: same bird, same session vs. same bird, different session

`processEncounterRow` upserts Encounters on `(bird_id, session_id)`. The test suite pins down
both sides of that key: re-processing the *identical* row updates in place (no duplicate,
correct for CSV re-imports); processing the *same bird on a different date* is a legitimately
new capture and must create a second row.

```mermaid
flowchart TB
    E1["Row: ring A123456, 12/06/2026, weight 10.5"] --> U1["upsert Encounters\n(bird_id=7, session_id=40)"]
    U1 --> ROW1["Encounters row #1\nweight = 10.5"]

    E2["Re-import same row, weight corrected to 12.3"] --> U2["upsert Encounters\n(bird_id=7, session_id=40)"]
    U2 -->|"same key → UPDATE"| ROW1B["Encounters row #1\nweight = 12.3 (unchanged row count)"]

    E3["Row: ring A123456, 13/06/2026 (new session)"] --> U3["upsert Encounters\n(bird_id=7, session_id=41)"]
    U3 -->|"different key → INSERT"| ROW2["Encounters row #2\nnew capture"]
```

Every table's `uniqueColumns` array also gets a direct check: a raw `.insert()` of a genuine
duplicate (bypassing `upsert` entirely) must fail with Postgres `23505` — proof the constraint
demon-import.ts relies on is actually still there.

supabase/__tests__/demon-import.test.ts
```diff
@@ -0,0 +1,712 @@
...
+			psqlCount(`SELECT COUNT(*) FROM "Species" WHERE species_name = '${speciesName}';`)
+		).toBe(1);
+
+		const speciesIds = psqlScalar(
+			`SELECT COUNT(DISTINCT species_id) FROM "Birds" WHERE ring_no IN ('DEMON-TEST-${suffix}-species-a', 'DEMON-TEST-${suffix}-species-b');`
+		);
+		expect(Number(speciesIds)).toBe(1);
+	});
+
+	it('rejects a raw duplicate species_name insert (bypassing upsert) with a unique-violation error', async () => {
```
wheresrhys: This is valuable, but scope creep for this test suite. Perhaps pull it out of here (as well as similar 'raw duplicate' tests) and instead create an issue to add another test suite which directly tests DB constraints for all tables

supabase/__tests__/demon-import.test.ts
```diff
@@ -0,0 +1,712 @@
...
+describe('demon-import — Birds uniqueness (ring_no)', () => {
+	let groupId: number;
+	let groupClient: SupabaseClient;
+
+	beforeAll(async () => {
+		groupId = createIsolatedGroup(`demon-import-${suffix}-birds`);
+		groupClient = await getAuthenticatedSupabaseClientForGroup(groupId);
+	});
+
+	it('upserting two rows with the same ring_no but a different species reuses the same Birds row and updates species_id', async () => {
```
wheresrhys: Interesting, this is not the desired behaviour and feels like a bug if this is possible. 

I know for a fact that the source data should not contain this sort of inconsistency. 

Leave this test out for now but add an issue to make species_id on ring_no immutable and to test for the correct behaviour when adding that change

supabase/__tests__/demon-import.test.ts
```diff
@@ -0,0 +1,712 @@
...
+	let groupClient: SupabaseClient;
+	let locationName: string;
+
+	beforeAll(async () => {
+		groupId = createIsolatedGroup(`demon-import-${suffix}-sessions`);
+		groupClient = await getAuthenticatedSupabaseClientForGroup(groupId);
+		locationName = `DemonImportLoc-${suffix}-sessions`;
+	});
+
+	it('upserting rows with the same visit_date/location but different session_type creates separate Session rows', async () => {
```
wheresrhys, line 410: Please also add tests for the other variants

- same location and session type, different date
- same visit date and session type, different location

supabase/__tests__/demon-import.test.ts
wheresrhys: Agreed — removed the raw-duplicate-insert-rejected-with-23505 tests from every uniqueness describe block in this file (Species, Locations, Sessions, Encounters — Birds' whole describe block went too, see the reply on the species_id comment). Filed #467 to track adding a dedicated DB-constraints test suite that covers all tables directly, independent of demon-import.ts.

supabase/__tests__/demon-import.test.ts
wheresrhys: Added both variants to the Sessions(visit_date, location_id, session_type) describe block: same location/session_type with a different date, and same visit_date/session_type with a different location — each asserting the two rows create separate Session rows, following the same isolation conventions as the rest of the file.

supabase/__tests__/demon-import.test.ts
wheresrhys: Good catch — removed this test (and, since it was the only other test in the describe block once the raw-dup test also went per the other comment, the whole "Birds uniqueness (ring_no)" describe block). Filed #468 to track making species_id immutable once set on a Birds row (by ring_no) — it'll need a schema-level trigger in supabase/schema/, so it's labelled db-migration — plus deciding how the import path should handle a same-ring_no/different-species row once that constraint exists, and testing the corrected behaviour.

Issue #467, opened 2026-08-13
Add dedicated test suite for DB uniqueness/constraint enforcement across all tables


wheresrhys:
## Background

PR #466 (#464) originally included, in each uniqueness-constraint describe block of
`supabase/__tests__/demon-import.test.ts`, a "raw duplicate insert rejected with `23505`"
test — bypassing `demon-import.ts`'s upsert logic and inserting directly via `.insert()`
to assert the underlying Postgres unique constraint itself rejects a literal duplicate.

Review feedback on that PR (https://github.com/wheresrhys/totf/pull/466#discussion_r3777870653)
flagged this as scope creep for a suite that's meant to test `demon-import.ts`'s upsert
behaviour, not the schema's constraints directly. Those tests were removed from that file.

## What this ticket covers

Add a new DB integration test suite (e.g. `supabase/__tests__/db-constraints.test.ts`)
that directly tests uniqueness (and other) constraints across all tables, independent of
any application code path. Suggested starting scope — one describe block per table,
each asserting a raw duplicate insert (bypassing any upsert helper) is rejected with
Postgres error code `23505`:

- `Species` (`species_name`)
- `Birds` (`ring_no`)
- `Locations` (`location_name`, `ringing_group_id`)
- `Sessions` (`visit_date`, `location_id`, `session_type`)
- `Encounters` (`bird_id`, `session_id`)

The removed tests in `supabase/__tests__/demon-import.test.ts` (see PR #466's history,
commit before the review-feedback fixup) are a ready-made reference for the pattern —
each raw-dup test there followed the same shape (insert once, expect `error).toBeNull()`;
insert again, expect `error?.code` to be `'23505'`).

Use the shared isolation helpers in `supabase/__tests__/test-isolation.ts`
(`randomTestSuffix`, `randomFutureDate`, `addDays`) so this suite is safe to run
concurrently with other worktrees against the shared local Supabase instance, per
CLAUDE.md's "DB integration tests" conventions.

## Suggested tests (USE algorithm)

**Structure** — one test per table's unique constraint (as listed above), each asserting
the raw duplicate insert is rejected with `23505`.

**Edge** — consider whether any table has more than one relevant constraint worth
covering (e.g. NOT NULL / FK violations), though the primary scope here is uniqueness
constraints matching `demon-import.ts`'s `uniqueColumns` per table.

## Out of scope

- Anything about `demon-import.ts`'s upsert/reuse-vs-create behaviour — that's already
[...]

PR #470, opened 2026-08-13
Make Birds.species_id immutable once set (#468)

### LINKED ISSUE #468: Make Birds.species_id immutable once set for a given ring_no
## Background

PR #466 (#464) originally included a test in
`supabase/__tests__/demon-import.test.ts` ("Birds uniqueness (ring_no)" describe block)
asserting that upserting two rows with the same `ring_no` but a *different*
`species_name` reuses the same `Birds` row and silently updates its `species_id` to the
second value.

Review feedback (https://github.com/wheresrhys/totf/pull/466#discussion_r3777884701)
flagged this as undesired/buggy behaviour, not a spec: a physical ring number should
never legitimately change species once assigned, so the source CSV data should never
contain this inconsistency, and the app currently masks a data-quality problem by
silently overwriting `species_id` instead of surfacing it. That test was removed from
the PR.

## What this ticket covers

1. **Schema change** (touches `supabase/schema/`, hence the `db-migration` label): make
   `Birds.species_id` immutable once set for a given row. `supabase/schema/schemas/public/tables/"Birds".sql`
   currently has no such protection — `species_id` is a plain `NOT NULL` FK column with
   only an UPDATE RLS policy (`group_birds_update`), no trigger guarding the column
   itself. Follow the existing trigger-function file convention in
   `supabase/schema/schemas/public/functions/` (e.g. `trg_encounters_refresh_bird_proven_age.sql`,
   `trg_add_bird_ringing_group_id.sql`) to add a `BEFORE UPDATE` trigger function (e.g.
   `trg_prevent_bird_species_id_change`) on `public."Birds"` that raises an exception
   if `NEW.species_id IS DISTINCT FROM OLD.species_id`, then wire it up in the table
   file per the declarative-schema workflow (`npm run db:diff` /
   `npm run db:schema:apply` — see CLAUDE.md's "Workflow for schema changes"). No
   backfill needed since this only constrains future writes.
2. **Decide + implement the correct import-time behaviour** once the constraint exists:
   `lib/demon-import.ts`'s `processEncounterRow`/`createUpserter` currently upserts
   `Birds` on `ring_no` and will attempt to write a changed `species_id` on a re-import
   with a different species — after this change that attempt will raise a DB error
   instead of silently succeeding. Decide how the import path should handle that (e.g.
   skip updating `species_id` on conflict / surface a validation error per row / log and
   continue) and implement it.
3. **Tests**:
   - A DB integration test (mirroring the removed one, but asserting the new *correct*
     behaviour) in `supabase/__tests__/demon-import.test.ts`'s "Birds uniqueness
     (ring_no)" describe block — upserting two rows with the same `ring_no` but a
     different `species_name` no longer silently changes `species_id`.
   - Consider a lower-level DB integration test asserting the trigger itself rejects a
     raw `UPDATE ... SET species_id = ...` on an existing `Birds` row (could live
     alongside the new constraint test suite from the sibling ticket, or here — use
     judgement at implementation time).

## Suggested tests (USE algorithm)

**Usual** — re-importing the same ring_no with the same species_name still works
(no-op update, unaffected by the new trigger).

**Structure** — one test for the trigger rejecting a direct `species_id` change; one
test for the import-path's new handling of a same-`ring_no`/different-`species_name` row.

**Edge** — updating a `Birds` row's `species_id` to the *same* value it already has
should not raise (only an actual change should be blocked).

## Also see

Related follow-up ticket on adding a dedicated DB-constraints test suite: #467.

wheresrhys:
> ⚠️ **Database migration — do not merge before pushing.** This PR's schema change must be
> deployed to prod (`npm run db:migration:push`, run via the `take-over` skill) before this
> PR is merged. Merging first triggers a Vercel prod deploy of code that expects a schema
> prod doesn't have yet.

Closes #468

## What this covers

Makes `Birds.species_id` immutable once set. A physical ring number belongs to one bird of one
species for life, so a re-imported CSV row reusing an existing `ring_no` with a *different*
species is a source-data error — not a legitimate update. Previously the import silently
overwrote `species_id`, masking the data-quality problem (see the removed test / review thread
in #466).

### Schema change (`db-migration`)

- New trigger function `public.trg_prevent_bird_species_id_change()` (guard-style, following the
  `trg_suppress_same_session_retrap` convention — plain `plpgsql`, no `SECURITY DEFINER`).
- New `BEFORE UPDATE` trigger `trigger_trg_prevent_bird_species_id_change` on `public."Birds"`.
- Raises `check_violation` (SQLSTATE `23514`) only when `NEW.species_id IS DISTINCT FROM
  OLD.species_id`. Setting `species_id` to its current value, and `INSERT`ing new birds, are
  unaffected. No backfill needed — this only constrains future writes.

The generated migration is pure DDL (`supabase/migrations/` is gitignored per repo convention —
regenerate at deploy time with `npm run db:schema:apply` / the `take-over` flow). For reference it
contains exactly:

```sql
CREATE FUNCTION public.trg_prevent_bird_species_id_change() RETURNS trigger LANGUAGE plpgsql AS $function$
BEGIN
  IF NEW.species_id IS DISTINCT FROM OLD.species_id THEN
    RAISE EXCEPTION 'species_id is immutable once set (ring_no=%): cannot change from % to %',
      OLD.ring_no, OLD.species_id, NEW.species_id
      USING ERRCODE = 'check_violation';
  END IF;
  RETURN NEW;
END;
$function$;
-- (+ GRANTs to anon/authenticated/service_role)
CREATE TRIGGER trigger_trg_prevent_bird_species_id_change BEFORE UPDATE ON public."Birds"
  FOR EACH ROW EXECUTE FUNCTION public.trg_prevent_bird_species_id_change();
```

### Import-path behaviour — no code change needed

Both `createUpserter` callers (`scripts/import-csv.ts` and `app/api/import/route.ts`) already
catch per-row errors and count the row as **failed** (the CLI also logs the error). The `Birds`
upsert on `ring_no` conflict now raises the trigger's exception when the species differs; that
propagates out of `processEncounterRow` into the existing per-row handlers, so the bad row is
surfaced as a failed record instead of silently overwriting `species_id`. This follows the
existing error-handling pattern rather than inventing a new one (per the ticket).

### Tests (`supabase/__tests__/triggers.test.ts`, new `Birds — species_id immutability trigger`
describe block)

Added to the existing DB-trigger integration suite (rather than the `supabase/__tests__/demon-import.test.ts`
/ `db-constraints.test.ts` files referenced in the ticket, since those live in the still-open #466
and unfiled #467 and would create a hard dependency / merge conflict):

- rejects a direct `species_id` change on an existing row (asserts `23514` + `species_id`
  preserved)
- allows setting `species_id` to the value it already has (only a real change is blocked)
- does not block creating a new Bird with a fresh `ring_no`
- import path (`createUpserter` on `ring_no`): re-importing the same species is a no-op update;
  re-importing a different species errors and preserves the original `species_id`

All 116 DB integration tests, 632 app tests, lint, and type-check pass locally; the push ran the
app suite + 91 E2E specs green.

## Incidental fix (separate commit)

`test: refresh fetchMistakes.alpha snapshot to seed-consistent order` — the committed
`fetchMistakes.alpha.json` fixture listed the `age` discrepancies with Robin first, but
`find_discrepencies` has no `ORDER BY` (seed-dependent order) and the mistakes page sorts rows by
species ascending. That left `mistakes/page.test.tsx > "renders ring_no as a link in each row"`
deterministically failing (it derives the expected `ring_no` from the fixture's raw order),
contradicting its sibling `"sorts rows by species ascending"` test — a pre-existing inconsistency
unrelated to the schema change. Regenerated the fixture via `db:generate-snapshots` so its rows are
in the species-sorted order the component renders; row contents are unchanged, only order.
Follow-up worth considering: give `find_discrepencies` a stable `ORDER BY` (or make the test
order-independent) so the fixture can't drift again.

- Placed the integration tests in `triggers.test.ts` to avoid depending on the unmerged #466/#467
  files the ticket named.
- Decided the correct import-time behaviour is "let the DB rejection fail the row through the
  existing per-row error handling" (surface, don't mask) rather than adding bespoke skip/validate
  logic.

### DIFF
--- a/supabase/__tests__/triggers.test.ts
+++ b/supabase/__tests__/triggers.test.ts
@@ -1,5 +1,5 @@
 /**
- * Integration tests for custom Postgres triggers on Encounters.
+ * Integration tests for custom Postgres triggers on Encounters and Birds.
  *
  * Requires local Supabase running and e2e seed data loaded:
  *   npm run db:start:local
@@ -12,6 +12,7 @@ import { describe, it, beforeAll, afterAll, expect } from 'vitest';
 import { execSync } from 'child_process';
 import { getAuthenticatedSupabaseClientForGroup } from '../../lib/group-auth';
 import { supabase } from '../../lib/supabase';
+import { createUpserter } from '../../lib/demon-import';
 import { randomTestSuffix } from './test-isolation';
 import type { SupabaseClient } from '@supabase/supabase-js';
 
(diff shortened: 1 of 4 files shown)


wheresrhys:
## Behaviour change: a re-import that changes a ring's species is now rejected, not silently applied

Worked example: `Birds` already holds `ring_no = AR0017 → species_id = 1 (Robin)`. A later CSV row
reuses `AR0017` but with `species_id = 58 (Wren)` — a data-quality error, since a ring belongs to
one bird of one species for life.

**Before → After** for that upsert (`onConflict: ring_no`):

```mermaid
flowchart TB
  subgraph OLD["Before — silent overwrite"]
    A1["Upsert AR0017, species 58"] --> A2["ON CONFLICT (ring_no)<br/>DO UPDATE"]
    A2 --> A3["Birds row: species 1 → 58"]
    A3 --> A4["row counted SUCCESSFUL<br/>bad data masked"]
  end
  subgraph NEW["After — rejected at the DB"]
    B1["Upsert AR0017, species 58"] --> B2["BEFORE UPDATE trigger<br/>NEW.species_id DISTINCT FROM OLD"]
    B2 --> B3["RAISE check_violation (23514)"]
    B3 --> B4["Birds row: species stays 1"]
    B3 --> B5["upsert throws → processEncounterRow rejects<br/>existing per-row catch counts row FAILED"]
  end
```

No import code changed: the trigger's exception rides the `createUpserter` throw path that
`scripts/import-csv.ts` and `app/api/import/route.ts` **already** catch per row — so the bad row
surfaces as a failed record instead of corrupting `species_id`.

## What still passes (only a real change is blocked)

The guard fires on `IS DISTINCT FROM` only, so ordinary re-imports and new rings are untouched:

| Write | `species_id` | Trigger |
|---|---|---|
| Re-import same ring, **same** species | 1 → 1 | no-op, allowed |
| Re-import same ring, **different** species | 1 → 58 | **RAISE 23514** |
| `INSERT` a brand-new ring | — (insert) | not a `BEFORE UPDATE`, allowed |

PR #482, opened 2026-08-19
Group slug resolver + route rename + href threading (#475, #476)

### LINKED ISSUE #475: Add lib/group-slug.ts resolver and rename app/group/[groupId] -> app/group/[groupSlug]
## Context

Part of the "Group slugs in URLs" effort (full plan: `.claude/plans/i-woudl-like-group-parallel-brook.md`, "Branch 3" section). We're replacing the numeric `RingingGroups.id` in cross-group page URLs (`/group/<id>/...`) with a human-readable `slug`. This ticket wires up the resolver and renames the dynamic route segment; it does **not** touch how links elsewhere in the app are built (that's a separate follow-up ticket).

Depends on **task 1** (adds nullable `slug text` column + unique constraint to `RingingGroups`, plus `lib/slugify.ts`) — this ticket needs the column to exist, but does *not* need task 2's `NOT NULL` enforcement/backfill, since resolution here just reads `slug` when present and treats an unresolved/null case as "not found."

1. **New `lib/group-slug.ts`**, following `lib/group-auth.ts`'s style (thin, no class, no existing cache utility — verified by grep there is no LRU cache anywhere in the repo despite CLAUDE.md's mention of one for JWT clients, so a plain `Map` is fine given the tiny number of groups):
   - `resolveGroupIdBySlug(slug: string): Promise<number | null>`
   - `resolveGroupSlugById(id: number): Promise<string | null>`
   - Each backed by its own module-scope `Map`, populated via a query against `RingingGroups` through `getAuthenticatedSupabaseClient()` (`lib/group-auth.ts`) — confirmed safe: `RingingGroups`' `SELECT` RLS policy is `USING (TRUE)` (fully open read), so any logged-in group's client can resolve any other group's slug/id.
   - **Only cache successful (non-null) resolutions.** Do not cache a miss — a `Map` has no eviction/TTL in this codebase, so caching `null` for an as-yet-unassigned slug/id would permanently 404 a group created or backfilled later in the same server process lifetime.
   - `resolveGroupSlugById` is not consumed by this ticket's own route-rename work — it's needed by the follow-up href-threading ticket (id→slug for the cookie-based path). Implement both now since they're a natural pair in the same file.

2. **`git mv app/group/[groupId] app/group/[groupSlug]`**, then update every file under it. Verified exact current shape of every file (`find "app/group/[groupId]" -type f`, 17 files):

   **12 page.tsx files — near-identical pattern**, `const { groupId } = await params; ... Number(groupId) ...` →`const { groupSlug } = await params; const viewedGroupId = await resolveGroupIdBySlug(groupSlug); if (viewedGroupId === null) notFound();`:
   - `effort/page.tsx`, `mistakes/page.tsx`, `pulli/page.tsx`, `resightings/page.tsx`, `retraps/page.tsx`, `sessions/page.tsx`, `species/page.tsx`, `species/[speciesName]/page.tsx`, `ticks/page.tsx` — all return `<XPage viewedGroupId={...} />` directly.
   - `session/[date]/page.tsx`, `session/[date]/site/[locationId]/page.tsx` — same shape but route through `BootstrapPageData` instead of a direct component call.
   - `page.tsx` (`CrossGroupHome`) — same shape, **plus** its existing `if (viewedGroupId === loggedInGroupId) redirect('/')` logic, which stays unchanged (just operates on the resolved id instead of `Number(groupId)`).

   Use `notFound()` from `next/navigation` (already used this way in `app/components/layout/BootstrapPageData.tsx:105`) when resolution returns `null`.

   **`layout.tsx` — no behavioural change needed.** It currently types `params: Promise<{ groupId: string }>` but never destructures or uses it (only checks the login cookie and redirects if absent). Just rename the type key to `groupSlug` for consistency with the folder rename; no resolver call needed here.

   **`session/_shared.tsx` — no code changes.** Verified by grep: it has zero references to `groupId`/`groupSlug` (it takes an already-resolved `viewedGroupId: number` as a plain function param/prop). It moves along with the `git mv` automatically. *(This deviates slightly from the original task sketch, which listed it as a file to "update" — inspection shows it needs no edits.)*

3. **Test-infra prerequisite found during drafting:** `vitest.setup.tsx`'s global `next/navigation` mock (used by all app tests) currently only stubs `useRouter`/`usePathname`/`useSearchParams`/`redirect` — it has no `notFound` export. Since `BootstrapPageData` (the only existing caller of `notFound()`) is itself globally mocked out, this gap has never surfaced. This ticket introduces the *first* `notFound()` calls in components that are **not** globally mocked, so add `notFound: vi.fn()` to that mock (mirrors how `redirect` is already stubbed) — otherwise any test exercising the null-resolution branch gets `TypeError: notFound is not a function`.

4. **Update the 3 existing test files** to match the new params shape:
   - `__tests__/layout.test.tsx` — rename `groupId: '1'` → `groupSlug: '...'` in all four fixtures (no assertions change; unused by the component).
   - `session/[date]/__tests__/page.test.tsx` — add `vi.mock('@/lib/group-slug', ...)` (hoisted, alongside the existing `@/lib/group-auth` mock) resolving to the numeric test group id; update `renderPage()`'s params to pass `groupSlug` instead of `groupId`; add a new test for the notFound branch (mock `resolveGroupIdBySlug` to resolve `null`, mock `notFound` to throw, assert `renderPage()` rejects — same pattern `__tests__/layout.test.tsx` already uses for `redirect`).
   - `session/[date]/site/[locationId]/__tests__/page.test.tsx` — same treatment.

5. **Two new test files**, to cover the new resolution/404 logic once each (not on every near-identical wrapper — see Out of scope):
   - `app/group/[groupSlug]/__tests__/page.test.tsx` (`CrossGroupHome`) — covers both its existing redirect-on-same-group behaviour and the new notFound-on-unresolved-slug behaviour, since it combines two conditionals.
   - `app/group/[groupSlug]/effort/__tests__/page.test.tsx` — one representative instance of the 9-file direct-component-call pattern (`effort`, `mistakes`, `pulli`, `resightings`, `retraps`, `sessions`, `species`, `species/[speciesName]`, `ticks` all share this exact shape).

## Out of scope

- Threading the slug down into `BootstrapPageData`/`PageComponent` props for href-building elsewhere in the app (`StatOutput.tsx`, `SessionHistoryCalendar.tsx`, `MonthSessions.tsx`, `StatsAccordion.tsx`/`StatsAccordionItem.tsx`, `SpStats.tsx`, `session/_shared.tsx`, `app/(routes)/page.tsx`, `app/(routes)/sessions/page.tsx`) — this is the follow-up ticket ("Branch 4"). It also owns wiring up `resolveGroupSlugById`'s actual call site.
- No redirect for legacy numeric URLs (`/group/123/...`) — they simply fail to resolve as a slug and 404 via `notFound()`. No legacy-URL support.
- Adding dedicated test files for the other 8 wrapper pages sharing the `effort/page.tsx` pattern — one representative test file is enough to prove the shared pattern; CLAUDE.md: "Don't unit test every simple component."
- Enforcing `slug NOT NULL` / backfilling existing rows (task 2, separately blocked/gated on manual prod backfill).

## Stack layers touched

- **Library/data-access** (new): `lib/group-slug.ts`.
- **Pages/routes** (renamed + edited): `app/group/[groupId]/**` → `app/group/[groupSlug]/**` (12 page.tsx, 1 layout.tsx; `session/_shared.tsx` moves only).
- **Tests**: `vitest.setup.tsx` (global mock addition), 3 existing test files updated, 2 new test files, 1 new lib unit test file.
- No DB schema, server actions, or shared components touched.

## Commit breakdown

1. Add `lib/group-slug.ts` (`resolveGroupIdBySlug`, `resolveGroupSlugById`, per-function `Map` cache, positive-only caching) + `lib/__tests__/group-slug.test.ts`.
2. `git mv app/group/[groupId] app/group/[groupSlug]`; update all 12 page.tsx files + `layout.tsx` to use `groupSlug`/`resolveGroupIdBySlug`/`notFound()`; add `notFound: vi.fn()` to `vitest.setup.tsx`'s `next/navigation` mock; update the 3 existing test files; add the 2 new test files (`page.test.tsx` for `CrossGroupHome`, `effort/__tests__/page.test.tsx` as the representative wrapper).

(Both commits land on one branch/PR — CLAUDE.md exempts "repetitive renaming" from the <400 LOC heuristic.)

## Tests (USE algorithm)

**`lib/__tests__/group-slug.test.ts`**
- `describe('resolveGroupIdBySlug')`
  - Usual: resolves a slug matching an existing `RingingGroups` row to its numeric id
  - Structure: returns `null` when no row matches the given slug
  - Edge: a second call with the same (resolved) slug does not issue a second Supabase query (cache hit)
  - Edge: a second call with the same (unresolved) slug re-queries Supabase (no negative caching)
- `describe('resolveGroupSlugById')`
  - Usual: resolves an id matching an existing row to its slug
  - Structure: returns `null` when no row matches the id, or the matching row's `slug` is `null`
  - Edge: a second call with the same (resolved) id does not issue a second Supabase query (cache hit)

**`app/group/[groupSlug]/__tests__/page.test.tsx` (`CrossGroupHome`)**
- `describe('slug resolves to a group id')`
  - Usual: renders `Home` with `viewedGroupId` set to the resolved numeric id, when it differs from the logged-in group's id
  - Structure: redirects to `/` when the resolved id equals the logged-in group's id (existing behaviour, unchanged)
- `describe('slug does not resolve')`
  - Edge: calls `notFound()` instead of rendering `Home` or redirecting

**`app/group/[groupSlug]/effort/__tests__/page.test.tsx`**
- Usual: renders `PayOffPage` with `viewedGroupId` set to the resolved numeric id when the slug resolves
- Edge: calls `notFound()` when the slug does not resolve

**Updates to existing test files** (no new describe blocks beyond what's listed):
- `__tests__/layout.test.tsx`: rename fixture key only, all 4 existing tests unchanged.
- `session/[date]/__tests__/page.test.tsx`: existing tests updated to mock `resolveGroupIdBySlug`; **add** "calls `notFound()` when the slug does not resolve" (Edge).
- `session/[date]/site/[locationId]/__tests__/page.test.tsx`: same as above.

## Verification

- `npm run test:nowatch`.
- Manual: `npm run db:start:local && npm run db:seed:e2e && npm run next:dev`, confirm `/group/<valid-slug>/...` renders and `/group/<invalid-or-numeric>/...` 404s (requires task 1 shipped with seeded slugs).

## Dependencies

- **Blocked by task 1** ("add nullable slug column + slugify util") — needs the `slug` column to exist.

### LINKED ISSUE #476: Thread group slug through href-building components (Branch 4)
## Context

Part of the "Group slugs in URLs" effort (full plan: group-slug/1..4 branches). Task 1 added a
`slug` column to `RingingGroups` + `lib/slugify.ts`. Task 2 will enforce `NOT NULL` on it. Task 3
adds `lib/group-slug.ts` (`resolveGroupIdBySlug`, `resolveGroupSlugById`) and renames
`app/group/[groupId]/` to `app/group/[groupSlug]/` across ~13 files, so cross-group routes carry
the slug in the URL instead of the numeric id.

This ticket is the one that actually makes the numeric id disappear from *rendered* URLs. A
numeric `RingingGroups.id` is used to build links back into the `/group/...` URL space from
several rendering components — not resolved once in one place — so those components need the
slug threaded down to them as a prop, since most of them are `'use client'` components that
cannot resolve it themselves (no server access, and the resolver in `lib/group-slug.ts` is
async/server-only).

**Why this is safe for the ~30 other `PageComponent` implementations that don't know about
`viewedGroupSlug`:** TypeScript structural typing allows a function whose declared parameter type
has *fewer* properties than what's actually passed to be assigned wherever the richer type is
expected (this is safe/contravariant for function parameters — the callee just ignores the extra
field). `BootstrapPageDataProps.PageComponent` will start receiving an extra `viewedGroupSlug`
field alongside the existing `viewedGroupId`; every other `PageComponent` implementation across
the app can keep destructuring only `{ params, data, viewedGroupId }` and will continue to
type-check and run unchanged. Verify this with `npm run type:check` (`tsc --noEmit` — see
`package.json`) once the `BootstrapPageData.tsx` change lands, before touching anything else.

## Scope discovered beyond the original plan text

The parent plan (Branch 4) named 9 files for the leaf-component swap. Tracing the actual call
graph (`grep`-ing every caller of the 7 named leaf components) surfaced **5 more files** that
must change too, or the build won't type-check once the leaf components' prop shape changes from
`viewedGroupId: number` to `viewedGroup: { id: number; slug: string }`:

- `app/components/YearSessions.tsx` — sits between `SessionHistoryCalendar` and `MonthSessions`
  in the prop chain (`SessionHistoryCalendar` → `YearSessions` → `MonthSessionsContent`). It
  currently forwards `viewedGroupId: number` through its own props and an internal `model` object;
  it builds no href itself but must forward `viewedGroup` instead for the chain to compile.
- `app/components/SpPage.tsx` — passes `viewedGroupId={viewedGroupId}` into `SpStats`. Once
  `SpStats`'s prop type changes, this call site breaks unless `SpPage` also receives/constructs
  `viewedGroup`.
- `app/(routes)/species/[speciesName]/page.tsx` — the `PageComponent` for this route is `SpPage`.
  Needs to accept `viewedGroupSlug`, bundle `{ id: viewedGroupId, slug: viewedGroupSlug }`, pass
  `viewedGroup` to `SpPage`.
- `app/group/[groupSlug]/species/[speciesName]/page.tsx` — cross-group wrapper around the above
  (same pattern as `app/group/[groupSlug]/page.tsx`); already has `speciesName`/`groupSlug` in
  its own URL params after task 3, so it must pass `viewedGroupSlug={groupSlug}` down.
- `app/(routes)/highlights/page.tsx` — passes `viewedGroupId={viewedGroupId}` into
  `StatsAccordion`. There is currently no `app/group/[groupSlug]/highlights/` route (grepped —
  doesn't exist), so this page is never rendered cross-group today, but it still must accept
  `viewedGroupSlug` and bundle `viewedGroup` purely to keep `StatsAccordion`'s new prop type
  satisfied. Because it's never rendered cross-group, `BootstrapPageData` always falls back to
  resolving the slug from the cookie's `loggedInGroupId`, which is correct here since
  `viewedGroupId` always equals `loggedInGroupId` on this route.

Two more files need updating for a reason the plan didn't call out: `_shared.tsx`'s href-building
components need the slug of the group being viewed, but `app/group/[groupSlug]/session/[date]/page.tsx`
and `.../site/[locationId]/page.tsx` call `BootstrapPageData` **directly** (with
`PageComponent={SessionSummary}`) rather than routing through `Home`/`SessionsPage` — they were
not in the plan's "pass `viewedGroupSlug={groupSlug}`" list, but if they're skipped,
`BootstrapPageData` would fall back to resolving the slug from the *logged-in* group's cookie
instead of the group actually being viewed, producing wrong links whenever viewing another
group's session cross-group:
- `app/group/[groupSlug]/session/[date]/page.tsx`
- `app/group/[groupSlug]/session/[date]/site/[locationId]/page.tsx`

Both must also pass `viewedGroupSlug={groupSlug}` directly into their own `BootstrapPageData` call.

### 1. `BootstrapPageData.tsx` contract change
- `BootstrapPageDataProps` gains `viewedGroupSlug?: string`.
- `LoadWithData` resolves it exactly like `viewedGroupId` today: prop override
  (`viewedGroupSlugProp`) takes precedence; otherwise derive via
  `resolveGroupSlugById(loggedInGroupId)` (`lib/group-slug.ts`, from task 3) using the cookie's
  numeric group id. Note `resolveGroupSlugById` returns `Promise<string | null>` — only call it
  when `loggedInGroupId` is truthy (mirror the existing `!viewedGroupId` early-return guard;
  don't add new error handling beyond what's already there for the equivalent numeric case).
- `PageComponent`'s declared prop type gains `viewedGroupSlug: string | null`, and
  `LoadWithData`'s JSX passes the resolved value through alongside `viewedGroupId`.

### 2. Page-level wiring (bundle `{ id, slug }` before handing to the render tree)
Each of these accepts `viewedGroupSlug` (server-resolved or URL-sourced) and constructs
`viewedGroup = { id: viewedGroupId, slug: viewedGroupSlug }` before passing it to the components
listed in §3:
- `app/(routes)/page.tsx` (`Home`) → `RecentSessions`/`SessionsByDay`
- `app/(routes)/sessions/page.tsx` (`SessionsPage`) → `SessionHistoryCalendar`
- `app/(routes)/species/[speciesName]/page.tsx` (`SpeciesPage`) → `SpPage`
- `app/(routes)/highlights/page.tsx` (`HighlightsPage`) → `StatsAccordion`
- `app/group/[groupSlug]/page.tsx` — passes `viewedGroupSlug={groupSlug}` into `Home`
- `app/group/[groupSlug]/sessions/page.tsx` — passes `viewedGroupSlug={groupSlug}` into
  `SessionsPage`
- `app/group/[groupSlug]/species/[speciesName]/page.tsx` — passes `viewedGroupSlug={groupSlug}`
  into `SpeciesPage`
- `app/group/[groupSlug]/session/[date]/page.tsx` — passes `viewedGroupSlug={groupSlug}` directly
  to its own `BootstrapPageData` call
- `app/group/[groupSlug]/session/[date]/site/[locationId]/page.tsx` — same, direct
  `BootstrapPageData` call

### 3. Leaf href-building components — swap `viewedGroupId: number` → `viewedGroup: { id: number; slug: string }`
Update the prop type, the `` `/group/${viewedGroupId}/...` `` template strings to
`` `/group/${viewedGroup.slug}/...` ``, and any other numeric usage (React `key`s) to
`viewedGroup.id`:
- `app/components/shared/StatOutput.tsx` (href ~line 59)
- `app/components/SessionHistoryCalendar.tsx` (href ~line 92, in `SessionsByDay`; also forwards
  to `YearSessions`)
- `app/components/YearSessions.tsx` (pure pass-through, no href of its own — forwards to
  `MonthSessionsContent`'s `model`)
- `app/components/MonthSessions.tsx` (forwards to `SessionsByDay`)
- `app/components/StatsAccordion.tsx` (has a `key={`${viewedGroupId}-...}`}` usage too, ~line 45)
- `app/components/StatsAccordionItem.tsx` (same `key` pattern, ~line 127; forwards to `StatOutput`)
- `app/components/SpStats.tsx` (href via `StatOutput`, ~line 160)
- `app/components/SpPage.tsx` (forwards to `SpStats`; does **not** forward to
  `SpIndividualsTab`/`SpNotableRetrapsTab`/`SpStatsHistoryTab`/`SpYearComparisonTab`/
  `SpWeightWingTab` — those keep taking `viewedGroupId: number` unchanged, they only use it for
  `.eq('ringing_group_id', ...)` queries, not hrefs — verified by grep, no `/group/${` in any of
  them)
- `app/group/[groupSlug]/session/_shared.tsx` — 4 href sites in `Locations` (2) and
  `SessionNavigation` (2). Do **not** touch `PageParams`, `fetchSessionData`, or
  `fetchAdjacentSessionDates` — those stay `viewedGroupId: number`-only (they're pure DB-query
  functions, out of scope per the exclusion below). Instead, give `SessionSummary` a new
  top-level `viewedGroupSlug: string | null` prop (arrives via the `PageComponent` contract from
  §1 — it already receives `params`, `data`, `viewedGroupId` as siblings, not nested inside
  `params`), and inside `SessionSummary` build
  `viewedGroup = { id: params.viewedGroupId, slug: viewedGroupSlug }` to pass down to `Locations`
  and `SessionNavigation`.

## Out of scope

Everything that only ever does `.eq('ringing_group_id', viewedGroupId)` for DB queries stays
completely untouched — this is the ~30 other files across `app/(routes)/*` and `app/actions/*`
(data-fetcher functions, RPC calls), plus the five `Sp*Tab` components listed above, plus
`_shared.tsx`'s `fetchSessionData`/`fetchAdjacentSessionDates`/`PageParams`. Confirmed by grep:
no other file has a `` `/group/${viewedGroupId}` `` (or equivalent) template string.

No schema changes (`supabase/schema/` untouched → no `db-migration` label).

## Dependencies

Blocked by task 3 (`group-slug/3-resolve-slug-in-route`) — needs `lib/group-slug.ts`
(`resolveGroupSlugById`) and the `app/group/[groupSlug]/` rename to exist first (this ticket's
diff assumes the renamed directory and `groupSlug` param names).

## Reuse

- `lib/group-slug.ts`'s `resolveGroupSlugById` (task 3) — same memoized-`Map` pattern as
  `resolveGroupIdBySlug`, already used by the renamed route files.
- `LoadWithData`'s existing `viewedGroupIdProp ?? loggedInGroupId` pattern in
  `app/components/layout/BootstrapPageData.tsx:87` — mirror exactly for `viewedGroupSlug`.
- The `{ id, slug }` bundling shape matches the plan's own stated design (Branch 4 section) —
  one object prop rather than two parallel scalar props, at the leaf-component boundary only.

## Stack layers touched

- Components (`app/components/*`) — 8 files (`StatOutput`, `SessionHistoryCalendar`,
  `YearSessions`, `MonthSessions`, `StatsAccordion`, `StatsAccordionItem`, `SpStats`, `SpPage`),
  plus `BootstrapPageData.tsx` (layout).
- Pages (`app/(routes)/*`, `app/group/[groupSlug]/*`) — 9 files.
- No server actions, no DB/schema, no RPC changes.
- Tests — component tests (`__tests__/` alongside each changed leaf component) and the existing
  page-level integration tests that already assert href output.

## Commit breakdown

Given this is a prop-shape change (not a rename) that must compile as one unit — every caller and
callee in the chain has to agree on the new shape simultaneously — this ships as a single PR.
Split into reviewable commits (exempt from the <400 LOC guideline per CLAUDE.md's "repetitive
renaming" carve-out — this is the same mechanical edit repeated ~17 times):

1. `BootstrapPageData.tsx`: add `viewedGroupSlug` prop + resolution + pass-through. Verify with
   `npm run type:check` that nothing else in the app breaks (proves the contravariance claim).
2. Leaf components (§3, 9 files): swap `viewedGroupId: number` → `viewedGroup: { id; slug }`,
   update hrefs/keys.
3. Page-level wiring (§2, 9 files): bundle `viewedGroup`, thread `viewedGroupSlug` through the
   two direct-`BootstrapPageData` session pages.
4. Update/add component + integration tests (see below).

## Tests (USE algorithm)

Most of these components have no dedicated unit test file today (verified: no test file exists
for `StatOutput`, `SessionHistoryCalendar`, `MonthSessions`, `StatsAccordion`, `SpPage` — only
`StatsAccordionItem.test.tsx`, `SpStats.test.tsx`, `YearSessions.test.tsx` exist, plus page-level
integration tests). Update existing tests' fixtures (`viewedGroupId={1}` →
`viewedGroup={{ id: 1, slug: 'alpha' }}`) and add href assertions where the component builds one
but no test currently checks it.

**`StatsAccordionItem.test.tsx` / `StatsAccordion` (update fixtures)**
- Structure: replace every `viewedGroupId={1}` fixture with `viewedGroup={{ id: 1, slug: 'alpha' }}`;
  assert the `key` prop no longer breaks (implicit via successful render — React would warn on
  duplicate/missing keys) and `StatOutput`'s rendered href uses `alpha`, not `1`.

**`YearSessions.test.tsx` (update fixtures)**
- Structure: same `viewedGroupId={1}` → `viewedGroup={{...}}` fixture swap; assert it still
  forwards correctly down to `MonthSessionsContent`/`SessionsByDay` (existing test presumably
  already renders far enough to hit a `StatOutput`/link — assert its `href` now contains `alpha`).

**`SpStats.test.tsx` (update fixtures)**
- Structure: same fixture swap; assert the "Top sessions" `StatOutput` links use
  `/group/alpha/session/...`.

**New: `StatOutput.test.tsx`** (`app/components/shared/__tests__/`)
- Usual: `temporalUnit="day"` with `viewedGroup={{id:1, slug:'alpha'}}` renders a link to
  `/group/alpha/session/<date>`.
- Structure: with a `location` prop, href includes `/site/<id>`; `link={false}` renders plain text,
  no anchor; non-`day` temporal units render no link regardless of `viewedGroup`.
- Edge: `temporalUnit === 'day'` and no `viewedGroup` → throws (existing behaviour, just confirm
  it still throws with the new prop name).

**New: `SessionHistoryCalendar.test.tsx`** (or extend if a suitable existing integration test
covers it — check `(routes)/sessions/__tests__/page.test.tsx` first, which renders this
component; add href assertions there if easier than a new file)
- Usual: a single-session day renders a `StatOutput` link to `/group/<slug>/session/<date>`.
- Structure: a multi-session day renders one link per location, each to
  `/group/<slug>/session/<date>/site/<locationId>`.

**New: `MonthSessions.test.tsx`**
- Usual: forwards `viewedGroup` through to `SessionsByDay` such that its rendered links use the
  slug (can assert via a rendered `StatOutput`/`NoPrefetchLink` href).

**New: `SpPage.test.tsx`** (or extend existing `Sp*Tab` test coverage / the
`(routes)/species/[speciesName]/__tests__/page.test.tsx` integration test)
- Usual: `SpStats`'s "Top sessions" links use `/group/<slug>/session/...` when `viewedGroup` is
  passed through from `SpeciesPage`.

**Update: `app/group/[groupSlug]/session/[date]/__tests__/page.test.tsx`** (renamed by task 3;
already asserts hrefs — `` `/group/${TEST_GROUP_ID}/session/...` ``)
- Usual: after this ticket, update those "Previous"/"Next" link assertions to expect
  `` `/group/${TEST_GROUP_SLUG}/session/...` `` (slug already available as a fixture per task 3's
  rename) — this is the strongest existing regression check that `_shared.tsx`'s
  `SessionNavigation` correctly threads `viewedGroupSlug`.
- Structure: add an assertion for the `Locations` badge-link href
  (`/group/<slug>/session/<date>/site/<locationId>`) — the multi-location branch isn't currently
  asserted at all; add coverage since this is exactly the kind of link this ticket changes.

**Edge cases (across the above)**
- `viewedGroupSlug` resolves to `null` (e.g. `resolveGroupSlugById` finds no row) — confirm this
  doesn't crash `BootstrapPageData` or downstream components; document the resulting (degraded but
  non-crashing) href as acceptable, since a group with no valid slug shouldn't reach this far once
  task 2's `NOT NULL` constraint is live — add one `BootstrapPageData`-level test for this rather
  than one per leaf component.

## Verification (manual)

```
npm run type:check
npm run test:nowatch
npm run db:start:local && npm run db:seed:e2e && npm run next:dev
```
Log in as a seeded group (e.g. Alpha), visit the sessions page and home page, confirm session
links read `/group/alpha/session/...` (not a numeric id), and confirm clicking through doesn't
404. Also visit a species page and the cross-group tree (`/group/beta/...` while logged in as
Alpha) to exercise the bundled-`viewedGroup` path end-to-end.

wheresrhys:
## Summary

This PR combines two chained tickets from the "Group slugs in URLs" effort into one branch/PR, since deploying #475 alone would break every existing cross-group href in the app (nothing built a slug-based link until #476 landed) — see the comment on #476 confirming the fold.

- **#475** — adds `lib/group-slug.ts` (`resolveGroupIdBySlug`, `resolveGroupSlugById`, each backed by a positive-only-cached `Map`) and renames `app/group/[groupId]/` to `app/group/[groupSlug]/` across all 17 files, resolving the slug to a numeric id server-side and 404ing via `notFound()` on an unresolved slug.
- **#476** — threads the resolved slug down through `BootstrapPageData` (new `viewedGroupSlug` prop/resolution) and every href-building leaf component (`StatOutput`, `SessionHistoryCalendar`/`SessionsByDay`, `YearSessions`, `MonthSessionsContent`, `StatsAccordion`, `StatsAccordionItem`, `SpStats`, `SpPage`, `session/_shared.tsx`'s `Locations`/`SessionNavigation`), so every `/group/<x>/...` link in the app now reads `/group/<slug>/...` instead of the numeric id.
- **E2E fix** (found during verification, not in either ticket's original scope): `e2e/authenticated/cross-group.spec.ts` and `cross-group-pages.spec.ts` built `/group/<id>/...` URLs directly — broken by the rename. Added `e2e/helpers/group-slugs.ts` (parallel to the existing `group-ids.ts` pattern, sourced from the real `slug` column via `global-setup.ts`) and switched both specs to build URLs from `alphaSlug`/`betaSlug`/`gammaSlug`.

## Note: pre-push hook bypassed once

This push used `--no-verify`, authorized by the user. The pre-push hook's `npm run test:nowatch` step reliably fails on `.claude/mcp/swarm-tools/lib/__tests__/state-file.test.ts` — a pre-existing bug in that test's repo-isolation (its `git rev-parse --git-common-dir` resolution doesn't isolate to its own temp repo, so it reads/writes the *real* `.claude/swarm-state.json`), completely unrelated to this PR's changes and reportedly being fixed separately by another agent. Everything else was verified directly, bypassing only that one hook step:

- `npx tsc --noEmit` — clean
- `npm run lint` — clean (no new warnings; 5 pre-existing unrelated ones)
- `npx vitest run` (full suite, excluding the broken swarm-tools file) — **744/744 passing**, 66 files
- `npx playwright test` on the two updated cross-group specs — **33/33 passing** against local Supabase
- `npx playwright test --grep-invert @mutates` (full safe E2E suite) — **91/91 passing**

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npm run lint`
- [x] `npx vitest run` — 744/744
- [x] `npx playwright test e2e/authenticated/cross-group.spec.ts e2e/authenticated/cross-group-pages.spec.ts` — 33/33
- [x] `npx playwright test --grep-invert @mutates` — 91/91

Closes #475
Closes #476


wheresrhys:
## Behaviour change: `/group/<id>/...` → `/group/<slug>/...`

### Why #475 and #476 had to ship together

`#475` alone renames the route segment and makes it resolve a **slug**, but nothing in the app was building slug-based links yet — every existing `href` (built by `StatOutput`, `SessionHistoryCalendar`, `session/_shared.tsx`, etc.) still interpolated the numeric `viewedGroupId`. Deploying #475 alone would have 404'd every cross-group link in the app the moment someone clicked it.

```mermaid
flowchart TB
    subgraph before["Before #476 (hypothetical: #475 alone)"]
        A1["User viewing Alpha (id 1, slug 'alpha')<br/>clicks a session link"] --> A2["Link href: /group/1/session/2024-05-10<br/>(component still uses numeric viewedGroupId)"]
        A2 --> A3["resolveGroupIdBySlug('1')<br/>no RingingGroups row has slug '1'"]
        A3 --> A4["404"]
    end

    subgraph after["After #476 (this PR, both tickets combined)"]
        B1["User viewing Alpha (id 1, slug 'alpha')<br/>clicks a session link"] --> B2["Link href: /group/alpha/session/2024-05-10<br/>(component now uses viewedGroup.slug)"]
        B2 --> B3["resolveGroupIdBySlug('alpha') → 1"]
        B3 --> B4["Session page renders"]
    end
```

### How a page gets its slug: prop override vs. cookie fallback

`BootstrapPageData` now resolves `viewedGroupSlug` the same way it already resolved `viewedGroupId` — a URL-sourced prop wins; otherwise it derives the *logged-in* group's own slug from the cookie. This lets the same `PageComponent` (e.g. `Home`, `SessionsPage`) work correctly whether it's rendered at `/` (own group) or `/group/beta/...` (cross-group), without knowing which case it's in.

```mermaid
flowchart TB
    Start["LoadWithData runs"] --> Prop{"viewedGroupSlug prop\npassed in?"}
    Prop -- "yes (cross-group route,\ne.g. /group/beta/sessions)" --> FromURL["Use the URL's own slug\n'beta'"]
    Prop -- "no (own-group route,\ne.g. /sessions)" --> Cookie{"loggedInGroupId\nfrom cookie?"}
    Cookie -- yes --> Resolve["resolveGroupSlugById(loggedInGroupId)"]
    Cookie -- no --> Null["null"]
    Resolve --> Found{"row found?"}
    Found -- yes --> CachedSlug["slug (cached for reuse)"]
    Found -- no --> Null
    FromURL --> Bundle["viewedGroup = { id, slug }\npassed to leaf components"]
    CachedSlug --> Bundle
    Null --> Bundle
```

A `null` slug (e.g. a group created before task 2's `NOT NULL` backfill) degrades gracefully — the affected link renders `/group/null/session/...` rather than crashing; covered by `BootstrapPageData.test.tsx`'s dedicated edge-case test.

app/(routes)/highlights/page.tsx
```diff
@@ -169,27 +169,33 @@ export async function fetchHighlightsData(
...
 	data,
-	viewedGroupId
+	viewedGroupId,
+	viewedGroupSlug
 }: {
 	data: StatsAccordionModel[];
 	viewedGroupId: number;
+	viewedGroupSlug: string | null;
 }) {
+	const viewedGroup = { id: viewedGroupId, slug: viewedGroupSlug };
```
wheresrhys: I thinked I'd prefer it if the viewedGroup object were created higher up in BootstrapPageData, do avoid this duplicated construction of the object in every page compnent

wheresrhys:
Done in 1836c3a — `viewedGroup = { id, slug }` is now constructed once in `BootstrapPageData`'s `LoadWithData` and passed as a single bundled prop, instead of each page rebuilding it. `PageComponent`'s type gained `viewedGroup` and dropped the now-redundant `viewedGroupSlug`; kept `viewedGroupId` since a couple of call sites (SpWeightWingTab, SessionHighlights) still need the raw id. Five page components simplified to just forward the prop. Tests/mocks updated to match.

app/components/layout/__mocks__/BootstrapPageData.tsx
```diff
@@ -65,6 +65,7 @@ export function BootstrapPageData<
 			params={params}
 			data={data}
 			viewedGroupId={1}
```
wheresrhys: You should now get rid of viewedGroupId now that this data is available via viewedGroup

app/components/layout/BootstrapPageData.tsx
```diff
@@ -11,6 +12,7 @@ export type BootstrapPageDataProps<DataType, PagePropsType, ParamsType> = {
 	loading?: React.ReactNode;
 	ttl?: number;
 	viewedGroupId?: number;
+	viewedGroupSlug?: string;
```
wheresrhys, line 15: In general aim to get rid of these props throughout the app. Instead pass around a viewedGroup object with id and slug properties. The exception is at the top level (mainly in this file I think, or potentially in a layout file within the ./group directory) where a group slug may be pulled from the params and a viewed group id derived from it, or alternatively for other routes where the logged in group id can be used to derive a group slug. 

I may be a little hazy on the detail of the mechanisms but the principle is that very early on in a page's lifecycle you should be able to derive a viewedGroup slug + id pair, so there is no need to pass around the individual props as they should pretty much always exist as a pair packaged in an object.

wheresrhys: Done. `BootstrapPageData` builds the `viewedGroup {id, slug}` pair once and now passes only that object to every `PageComponent`; the redundant standalone `viewedGroupId` prop is removed from the `PageComponent` contract (both the real component and the mock). So nothing below the top level receives the individual scalars any more. The individual `viewedGroupId?`/`viewedGroupSlug?` inputs remain only here at the top-level boundary (and the `app/group/[groupSlug]` route layer) where they are derived from params/cookie — the exception you described. Removing the prop from the shared contract cascaded (via the type checker) to a handful of pre-existing PageComponents — `SpPage`, `PayOffPageContent`, `BirdSummary`, `SppStatsTable`, `ControlsPage`, `RingSequencesPage` — which now read `viewedGroup.id`; behaviour is unchanged.

app/components/layout/__mocks__/BootstrapPageData.tsx
wheresrhys: Done — the mock no longer passes `viewedGroupId={1}`; it passes only `viewedGroup={{ id: 1, slug: ... }}`, mirroring the real `BootstrapPageData` now that the `PageComponent` contract has dropped the standalone `viewedGroupId` prop (id is read from `viewedGroup.id`).

app/(routes)/highlights/page.tsx
wheresrhys: This one was already resolved by the current head commit (1836c3a), which moved `viewedGroup` construction up into `BootstrapPageData` instead of building it per page. The follow-up change reinforces it: `viewedGroup` is now the single object passed to every `PageComponent`, and the parallel `viewedGroupId` prop has been removed from the contract, so there is no per-page reconstruction of the object anywhere.

app/group/[groupSlug]/sessions/page.tsx
```diff
@@ -0,0 +1,16 @@
...
+export default async function CrossGroupSessionsPage({
+	params
+}: {
+	params: Promise<{ groupSlug: string }>;
+}) {
+	const { groupSlug } = await params;
+	const viewedGroupId = await resolveGroupIdBySlug(groupSlug);
+	if (viewedGroupId === null) notFound();
+	return (
+		<SessionsPage viewedGroupId={viewedGroupId} viewedGroupSlug={groupSlug} />
```
Comment (wheresrhys) [inline app/group/[groupSlug]/sessions/page.tsx:14]: Here's another example of separate props. To be clear, wherever possible bundle them into a single viewedGroup: {id, slug} prop

Issue #491, opened 2026-08-19
swarm §1 PR-maintenance workers only fetch+merge origin/main when GitHub reports a conflict — feedback-only PRs stay stale


wheresrhys:
## Context

Issue #485 documents a failure mode: a git worktree/branch cut before an infra fix lands on `origin/main` keeps re-triggering an already-fixed bug, because nothing nudges it to pick up the fix. The concrete incident was `.claude/mcp/swarm-tools/lib/state-file.ts`'s `GIT_DIR`-leak bug (fixed in PR #483) — a branch cut before #483 merged kept corrupting the shared `.claude/swarm-state.json` on every `git push` (via the pre-push hook, which runs the swarm-tools test suite) until manually merged with `origin/main`. A separate, already-in-progress fix (different worktree) adds a `.husky/pre-push` guard for the *push-time* case — that work is out of scope here and must not be touched.

This ticket fixes the same failure class at an earlier point: `.claude/skills/swarm/SKILL.md` §1 ("Maintain open PRs"). Its worker prompt currently reads:

> 1. **Conflicts first** — if `CONFLICTING`, `git fetch origin` then `git merge origin/main` ...
> 2. **Then feedback** — summarise outstanding feedback, make the changes, run tests, reply to the reviewer via `gh pr comment <n>`.

The fetch+merge is gated on the PR's `mergeable` status being `CONFLICTING`. GitHub's mergeable check only detects textual conflicts, not staleness — a PR can be many commits behind `origin/main` on unrelated files (e.g. a swarm-tools infra fix) and still report `MERGEABLE`. So for a feedback-only PR (`reason: "feedback"`, the common case), the worker never fetches or merges `origin/main`. It reuses the existing worktree exactly as left, edits, and **runs tests** (step 2) against whatever `origin/main` looked like when the worktree/branch was cut — silently reintroducing a since-fixed infra bug, with no push-time guard catching it until step 3's later `push`.

## Scope & acceptance criteria

- `.claude/skills/swarm/SKILL.md` §1's worker prompt unconditionally runs `git fetch origin` + `git merge origin/main` as the first step for every maintenance worker, regardless of `mergeable` status — before any feedback-driven editing or test-running.
- Conflicts produced by that merge (whether GitHub already flagged `CONFLICTING` or the fetch surfaces one GitHub's cache hadn't caught yet) are resolved using the same "honour both sides' intent, run tests to prove the merge is sound" language already present.
[...]

Issue #484, opened 2026-08-19
swarm-state.json schema drift crashes swarm_plan_batch / swarm_state_append


wheresrhys:
## Bug

`.claude/swarm-state.json` is supposed to be a bare JSON array of `SwarmWorkerEntry` (per `lib/state-file.ts`'s `readState`/`writeState`, which do `JSON.parse(raw) as SwarmWorkerEntry[]` and `JSON.stringify(entries, ...)` with no top-level wrapper). At session start today the file on disk held `{"workers": [...24 bogus entries...]}` instead — an object, not an array. That shape mismatch isn't caught anywhere:

- `readState` blindly casts `JSON.parse(raw) as SwarmWorkerEntry[]` with no runtime shape check, so a wrapped object round-trips silently as "valid" until a caller iterates it.
- `swarm-plan-batch.ts:246-247` does `const running = await listState(); for (const entry of running)` — with an object instead of an array this throws `TypeError: running is not iterable` at runtime, which is exactly what `mcp__swarm-tools__swarm_plan_batch` surfaced.
- `swarm_state_append` (via `withStateLock`) hit a related failure (`entries is not iterable`) trying to mutate the same malformed file.

Repro: `echo '{"workers":[]}' > .claude/swarm-state.json` then call `swarm_plan_batch` or `swarm_state_append` — both throw instead of treating it as empty state or reporting a clear schema error.

## Origin

Unclear how the file first got the `{"workers":[...]}` wrapper with 24 duplicate/fake entries (`agent-0`..`agent-7` each repeated ~3x, `worktreePath: "/tmp/example"`, `branch: "feature/1-example"`, all `startedAt: "2026-01-01T00:00:00.000Z"`) — looks like leftover test fixture data written to the real path rather than a test tmpdir, or a stale schema from before the array-shaped rewrite. Worth checking whether `tools/__tests__/*.test.ts` or `lib/__tests__/state-file.test.ts` ever point their fixtures at the real project path instead of a temp dir.

## Suggested fix

1. In `readState` (`lib/state-file.ts`), validate the parsed JSON is actually an array before returning it — on a non-array shape, throw a clear, named error (e.g. `SwarmStateSchemaError: expected an array, got object — see .claude/swarm-state.json`) instead of letting every downstream `.map`/`for...of` fail with a generic "is not iterable".
2. Audit the test suite for any fixture that writes to the real repo path instead of an isolated tmpdir — that's the likeliest source of the corruption.
[...]

PR #492, opened 2026-08-19
fix: auto-merge origin/main before push when swarm-tools is stale

### LINKED ISSUE #485: Stale worktrees repeatedly hit swarm-tools state-file test flake already fixed on main
## Bug / process gap

The `.claude/mcp/swarm-tools` state-file test suite has now caused pre-push failures on the same branch (`feature/475-add-lib-group-slug-ts` / ticket #475) twice, in two separate sessions, for what turned out to be one root cause: **a git worktree/branch cut before an infra fix lands on `main` never automatically picks it up**, so every worker still on an old base keeps re-hitting a bug that's already fixed elsewhere.

### Timeline

1. An earlier #475 worker root-caused the original bug: `resolveProjectRoot()` in `lib/state-file.ts` (and `state-file.test.ts`'s own `git init` call) shell out to `git` without stripping `GIT_DIR`/`GIT_WORK_TREE`/`GIT_INDEX_FILE`. Husky sets `GIT_DIR` when invoking `.husky/pre-push`, so any test run through that hook inherits it — `git rev-parse --git-common-dir` (and `git init`) then resolve the *real* repo instead of the test's isolated temp dir, and `state-file.test.ts` reads/writes the live, shared `.claude/swarm-state.json` instead of its own fixture.
2. Fixed in commits `5e84ff4` / `32574a0`, merged to `main` via **PR #483** ("fix: state-file test isolation escapes to real swarm-state.json under GIT_DIR").
3. Before that fix reached every open branch, the escape had already corrupted the real `.claude/swarm-state.json` with a wrapped `{"workers": [...24 bogus entries...]}` shape — literally `state-file.test.ts`'s own `makeEntry()` fixtures (`agent-0..agent-7`, `worktreePath: "/tmp/example"`, `branch: "feature/1-example"`). **Issue #484** documents that corruption and the resulting `TypeError: ... is not iterable` crashes in `swarm_plan_batch` / `swarm_state_append`.
4. Today, in a separate session on the *same* `feature/475-add-lib-group-slug-ts` branch (after PR #483 had already merged), I hit the identical `state-file.test.ts` flake again on `git push` — 4 consecutive attempts, all failing with `TypeError: current/entries is not iterable`. Cause: this branch was cut before PR #483 merged and was never rebased/merged with `main`, so it was still running the unpatched `state-file.ts`/`state-file.test.ts`, which duly re-corrupted the real `.claude/swarm-state.json` (found as `{"workers":[]}` on disk — the wrong schema, mid-incident). The isolated test file passed 6/6 on its own; only the full suite, invoked via the real pre-push hook (which sets `GIT_DIR`), failed. Merging `origin/main` into the branch (to pick up PR #483) fixed it immediately — the next `git push` passed cleanly through app tests, lint, type-check, and the full E2E suite.

### Root cause (the part not yet fixed)

PR #483 fixed the underlying `GIT_DIR` leak, and #484 tracks hardening `readState`/`writeState` against schema drift. But nothing stops a worktree that predates a swarm-tools infra fix from continuing to hit it: `swarm`/`take-over`/an ordinary `git push` don't check whether `.claude/mcp/swarm-tools/**` on the branch is behind `origin/main` before running the test suite that exercises it. Any worker whose branch was cut before a fix like #483 lands will keep reproducing the exact same failure — and, worse, keep re-corrupting the *shared* `.claude/swarm-state.json` for every other concurrently-running worker — until it happens to merge/rebase onto `main`.

### Suggested fix

Add a staleness guard (e.g. alongside `scripts/e2e-select-suite.sh`, in the pre-push hook itself, or as a `swarm`/`take-over` pre-flight step) that detects when the current branch is behind `origin/main` specifically on `.claude/mcp/swarm-tools/**`, and either:
- warns loudly and suggests `git merge origin/main`, or
- auto-merges `origin/main` before running the swarm-tools test suite (mirroring `take-over`'s existing "merges in `origin/main` so the branch/PR is current" step).

This is narrower than, and complementary to, #484's schema-validation suggestion (which guards against corruption *after* it happens); this issue is about preventing the repeat *cause* — stale worktrees never being nudged to pick up already-merged swarm-tooling fixes.

### Impact

- Wasted worker time: 4 failed push attempts in one session alone before diagnosing.
- Corrupts `.claude/swarm-state.json` for every other concurrently-running worker each time it recurs (per #484's incident).
- Will keep recurring for any branch/worktree cut before a swarm-tools fix, for any *future* swarm-tools infra fix, not just this one.

### Related

- PR #483 (merged) — the `GIT_DIR` isolation fix itself.
- Issue #484 (open) — schema-drift hardening for `readState`/`writeState`.
- Issue #478 (open) — tracks remaining swarm-tools MCP work generally.

wheresrhys:
## Summary

- Adds `scripts/check-swarm-tools-staleness.sh`, run first in `.husky/pre-push`: if the current branch is behind `origin/main` specifically on `.claude/mcp/swarm-tools/**`, it auto-merges `origin/main` before the rest of the hook (lint/typecheck/tests/e2e) runs.
- Prevents the failure mode documented in #485: a branch cut before a swarm-tools infra fix lands on `main` keeps re-running the unpatched code and re-corrupting the shared `.claude/swarm-state.json` on every push, until someone manually merges `origin/main` by hand.
- On a clean merge: prints a confirmation, continues the push. On a conflicting merge: aborts the merge (`git merge --abort`), prints a message telling the dev to resolve manually, exits 1 (blocks the push rather than leaving a half-merged tree).
- No-op on `main` itself or when the branch is already current on that path.

## Related

- Closes #485
- PR #483 (merged) — the underlying `GIT_DIR` isolation fix this guard protects against recurring
- Issue #484 (open) — separate schema-drift hardening for `readState`/`writeState`
- Issue #491 (opened during this work) — a related gap found in `swarm`'s PR-maintenance worker, which can still run tests against a stale `origin/main` when a PR needs feedback-only maintenance (no textual conflict to trigger its existing merge step)

## Test plan

- [x] `npm run qa` (lint + typecheck + 717 app tests) passes
- [x] Manually verified in a scratch clone: stale branch fast-forward-merges cleanly (exit 0), already-current branch no-ops (HEAD unchanged, exit 0), conflicting stale branch aborts the merge and exits 1 with a clean working tree
- [x] Real `git push` on this branch ran the hook end-to-end (script + lint + typecheck + tests + full e2e suite), all green

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,3 +1,4 @@
+./scripts/check-swarm-tools-staleness.sh
 npm run lint
 npx tsc --noEmit
 npm run test:nowatch
--- a/scripts/check-swarm-tools-staleness.sh
+++ b/scripts/check-swarm-tools-staleness.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+set -e
+
+# Every git worktree of this repo shares one common .git dir, so a branch cut before an infra
+# fix lands on .claude/mcp/swarm-tools keeps re-running the unpatched code (and re-corrupting the
+# shared .claude/swarm-state.json) every time its test suite runs through this hook — see issue
+# #485. Auto-merge origin/main before running tests when the branch is behind on that path.
+
+CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+if [ "$CURRENT_BRANCH" = "main" ]; then
+	exit 0
+fi
+
+MERGE_BASE=$(git merge-base HEAD origin/main)
+STALE=$(git diff --name-only "$MERGE_BASE" origin/main -- .claude/mcp/swarm-tools/)
+
+if [ -z "$STALE" ]; then
+	exit 0
+fi
+
+echo "check-swarm-tools-staleness: branch is behind origin/main on .claude/mcp/swarm-tools/** — auto-merging origin/main before running tests"
+
+if ! git merge origin/main -m "Merge origin/main: pick up swarm-tools fixes before push"; then
+	git merge --abort
+	echo "check-swarm-tools-staleness: auto-merge failed (conflicts) — resolve manually with 'git merge origin/main' and retry push" >&2
+	exit 1
+fi
+
+echo "check-swarm-tools-staleness: merged origin/main successfully"


wheresrhys:
## Stale-branch pushes now self-heal instead of flaking

Before this PR, a branch cut before a `.claude/mcp/swarm-tools` fix (e.g. PR #483's `GIT_DIR` isolation fix) kept re-running the unpatched test code on every `git push` — this actually happened on `feature/475-add-lib-group-slug-ts`, which flaked 4 times in a row before someone noticed and ran `git merge origin/main` by hand.

```mermaid
flowchart TB
    subgraph before["Before — no guard"]
        direction TB
        A1["git push (branch predates PR #483)"] --> A2["pre-push: lint → tsc → tests"]
        A2 --> A3["state-file.test.ts hits GIT_DIR leak"]
        A3 --> A4["corrupts shared .claude/swarm-state.json"]
        A4 --> A5["push fails — repeat 4x until dev manually merges origin/main"]
    end

    subgraph after["After — this PR"]
        direction TB
        B1["git push"] --> B2["check-swarm-tools-staleness.sh (runs first)"]
        B2 --> B3{"branch behind origin/main\non .claude/mcp/swarm-tools/**?"}
        B3 -->|no| B4["no-op — continue to lint/tsc/tests"]
        B3 -->|yes| B5{"git merge origin/main"}
        B5 -->|clean| B6["merged — continue to lint/tsc/tests, now against patched code"]
        B5 -->|conflict| B7["git merge --abort — clean tree,\nexit 1 with 'resolve manually' message"]
    end
```

Clean-merge and no-op cases push through automatically; a genuine conflict now fails fast with a clear message instead of a confusing downstream test flake.

PR #493, opened 2026-08-19
fix: validate swarm-state.json shape with a clear named error

### LINKED ISSUE #484: swarm-state.json schema drift crashes swarm_plan_batch / swarm_state_append

wheresrhys:
## Summary

- `readState` (`lib/state-file.ts`) previously cast `JSON.parse`'s result straight to `SwarmWorkerEntry[]` with no runtime check — a malformed file round-tripped as "valid" until a caller iterated it and hit a generic `TypeError` (`entries.filter is not a function`, `running is not iterable`).
- Validates with a `zod` schema instead, throwing a named `SwarmStateSchemaError` with a specific path/message for both top-level shape mismatches (object instead of array) and per-entry problems (missing/mistyped fields).
- The same schema (`swarmWorkerEntrySchema`, moved to `lib/state-file.ts`) now backs the `swarm_state_*` tools' output typing too, replacing a duplicate hand-written zod object that previously lived in `tools/swarm-state.ts`.
- Audited the test suite per the ticket's second suggestion: the only fixture that could write to the real repo path was already fixed in #483 (`state-file.test.ts` isolates via an `mkdtemp`'d temp repo with `GIT_DIR` stripped) — no other source of corruption found.

## Verification

Ran the fixed `readState` against the actual corrupted `.claude/swarm-state.json` on disk (`{"workers":[]}`):

```
name: SwarmStateSchemaError
message: .../.claude/swarm-state.json does not match the expected SwarmWorkerEntry[] shape: (root): Invalid input: expected array, received object
```

vs. the previous generic `entries.filter is not a function`.

## Related

- Closes #484
- Issue #485 / PR #492 — the other half of the swarm-state corruption story (stops branches from re-hitting an already-fixed swarm-tools bug on push)

## Test plan

- [x] `npm run qa` — lint + typecheck + all 723 app tests (6 new: object-wrapper, missing key, wrong type, non-array/non-object top level, empty array accepted, unknown extra keys tolerated)
- [x] Manually verified against the real corrupted state file (see above)

### DIFF
--- a/.claude/mcp/swarm-tools/lib/__tests__/state-file.test.ts
+++ b/.claude/mcp/swarm-tools/lib/__tests__/state-file.test.ts
@@ -4,7 +4,7 @@ import { promises as fs } from 'fs';
 import os from 'os';
 import path from 'path';
 import { execa } from 'execa';
-import { withStateLock, listState, type SwarmWorkerEntry } from '../state-file';
+import { withStateLock, listState, SwarmStateSchemaError, type SwarmWorkerEntry } from '../state-file';
 
 function makeEntry(overrides: Partial<SwarmWorkerEntry> = {}): SwarmWorkerEntry {
 	return {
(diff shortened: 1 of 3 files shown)


wheresrhys:
## Malformed state file now fails loudly instead of confusingly

`readState` used to trust `JSON.parse`'s result was already `SwarmWorkerEntry[]` — a wrapped or malformed file round-tripped as "valid" until some caller iterated it deep inside `swarm_plan_batch` or `swarm_state_append`, surfacing a generic JS error with no clue what was actually wrong.

```mermaid
flowchart TB
    subgraph before["Before"]
        direction TB
        A1["swarm_plan_batch reads .claude/swarm-state.json"] --> A2["JSON.parse cast straight to SwarmWorkerEntry[]"]
        A2 --> A3["isSoloRunCurrentlyActive: for (const entry of running)"]
        A3 --> A4["running is {workers: []}, not an array"]
        A4 --> A5["TypeError: running is not iterable"]
    end

    subgraph after["After"]
        direction TB
        B1["swarm_plan_batch reads .claude/swarm-state.json"] --> B2["zod: swarmStateSchema.safeParse(parsed)"]
        B2 --> B3{"matches SwarmWorkerEntry[]?"}
        B3 -->|yes| B4["return entries, tool proceeds normally"]
        B3 -->|no| B5["SwarmStateSchemaError: '(root): Invalid input: expected array, received object'"]
    end
```

Verified against the actual corrupted file on disk (`{"workers":[]}`) — before this fix it threw `entries.filter is not a function` depending on the call path; after, every caller gets the same specific, named error pointing at the exact mismatch.

PR #495, opened 2026-08-19
Add lib/group-slug.ts resolver + compute viewedGroup for /group route pages

wheresrhys:
## Summary

Step 1 of the "Group slugs in URLs" effort's re-split (supersedes the scope of #482). This ticket only computes and *carries* a `{ id, slug }` object through the `/group/[groupId]/...` wrapper layer and widens the type of the route pages they call — it does not change any rendering/routing/caching behaviour. Tickets 2 and 3 will consume it.

- Adds `lib/group-slug.ts` (`resolveGroupIdBySlug` / `resolveGroupSlugById`, each backed by a module-scope, cache-only-on-hit `Map`) and `lib/__tests__/group-slug.test.ts`, reused verbatim from the closed PR #482 (branch `feature/475-add-lib-group-slug-ts`). Also exports a shared `ViewedGroup` type (`{ id: number; slug: string | null }`), used by every call site below rather than repeating the inline shape 28 times.
- Updates the 13 files under `app/group/[groupId]/**` (12 `page.tsx` + `session/_shared.tsx`) that parse the numeric `groupId` param and call the corresponding `app/(routes)/*` page: each now computes `const viewedGroup = { id: viewedGroupId, slug: await resolveGroupSlugById(viewedGroupId) }` and passes it as a new, additional `viewedGroup` prop alongside the existing `viewedGroupId`.
- Widens the prop type of the 15 `app/(routes)/*/page.tsx` files these wrappers call to accept an optional `viewedGroup?: ViewedGroup` prop — deliberately not destructured or consumed anywhere yet (TypeScript's excess-property check on JSX requires the type to exist even though nothing reads it).

## Assumptions / notes

- `lib/group-slug.ts` assumed `RingingGroups.slug` already exists in the schema (it does — the DB-migration steps of the "Group slugs" effort landed separately).
- Fixed the two pre-existing session wrapper test suites (`app/group/[groupId]/session/[date]/__tests__/page.test.tsx` and its `site/[locationId]` sibling), whose mocked Supabase clients queued `.from()` responses positionally. The extra `RingingGroups` slug lookup broke that ordering — and, because `resolveGroupSlugById`'s cache is module-scoped, later tests reusing the same group id skip the query entirely, which made simple index-shifting the wrong fix. Reworked the mocks to dispatch on table name instead, so they're robust regardless of whether a given test hits the cache.
[...]

Issue #488, opened 2026-08-19
Thread viewedGroup into PageComponents; drop redundant viewedGroupId


wheresrhys:
## Context

Part of the "Group slugs in URLs" effort (#472), step 3 of 3 — depends on #487. This is where `viewedGroupId` finally disappears from the `PageComponent` boundary and from the `/group` wrapper layer's calls.

## Scope

- `app/components/layout/BootstrapPageData.tsx`: `PageComponent`'s prop type changes `viewedGroupId: number` → `viewedGroup: { id: number; slug: string | null }`; `LoadWithData`'s render passes `viewedGroup={viewedGroup}` instead of `viewedGroupId={viewedGroupId}`. Update the mock's `PageComponent` call the same way (`viewedGroup={{ id: 1, slug: 'alpha' }}` instead of `viewedGroupId={1}`).
- Update the `PageComponent` implementations that actually destructure `viewedGroupId` today (confirmed by grep — the rest receive it structurally but never read it, so they need no change):
  - Inline in their route file: `HomePageContent` (`app/(routes)/page.tsx`), `PayOffPageContent` (`effort/page.tsx`), `ListAllSessions` (`sessions/page.tsx`), `BirdSummary` (`bird/[ring]/page.tsx`), `HighlightsPageContent` (`highlights/page.tsx`)
  - Imported components: `SpPage.tsx`, `SppStatsTable.tsx`, `ControlsPage.tsx`, `RingSequencesPage.tsx`
  - Each swaps its `viewedGroupId: number` prop for `viewedGroup: { id: number; slug: string | null }` and uses `viewedGroup.id` wherever it previously used `viewedGroupId` (including when passing the plain numeric id further down to its own children, e.g. `SpPage` → `SpStats`/`SpYearComparisonTab`/etc., `ListAllSessions` → `SessionHistoryCalendar`). Those grandchildren keep receiving a plain `viewedGroupId: number` as before — this ticket does not cascade the object shape past the `PageComponent` boundary.
  - No change needed: `ListTicks`, `ListMistakes`, `ListResightings`, `ListNotableRetraps`, `ListPulliEncounters`, `BirdSearchResults` (none destructure `viewedGroupId`).
- Drop the now-redundant `viewedGroupId={viewedGroupId}` prop from the 13 `/group/[groupId]/**` wrapper files added in #486 — they pass only `viewedGroup` from here on.
[...]

Issue #486, opened 2026-08-19
Add lib/group-slug.ts resolver + compute viewedGroup for /group route pages


wheresrhys:
## Context

Part of the "Group slugs in URLs" effort (#472). Supersedes the scope of #482 by splitting it into three small, sequential PRs instead of one sprawling one — this is step 1 of 3. Reuses `lib/group-slug.ts` and its test file verbatim from PR #482 (branch `feature/475-add-lib-group-slug-ts`) rather than re-deriving them.

This ticket only computes and *carries* a `{ id, slug }` object through the `/group/[groupId]/...` wrapper layer and widens the type of the route pages they call. It does not change any rendering/routing/caching behaviour — the object is inert until tickets 2 and 3 land.

## Scope

- Add `lib/group-slug.ts` — `resolveGroupIdBySlug(slug)` / `resolveGroupSlugById(id)`, each backed by a module-scope, cache-only-on-hit `Map` (verbatim copy from `feature/475-add-lib-group-slug-ts`).
- Add `lib/__tests__/group-slug.test.ts` (verbatim copy — covers resolve/miss/cache-hit for both directions).
- Update the 13 files under `app/group/[groupId]/**` (`page.tsx` × 12 + `session/_shared.tsx`) that currently parse the numeric `groupId` param and call the corresponding `app/(routes)/*` page: compute `const viewedGroup = { id: viewedGroupId, slug: await resolveGroupSlugById(viewedGroupId) }` and pass it as a **new**, additional `viewedGroup={viewedGroup}` prop — alongside the existing `viewedGroupId={viewedGroupId}`, which keeps flowing unchanged.
- Widen the prop type of each of the 15 `app/(routes)/*/page.tsx` files (the ones these wrappers call: `page.tsx`, `ticks/`, `effort/`, `mistakes/`, `resightings/`, `retraps/`, `species/`, `species/[speciesName]/`, `controls/`, `sessions/`, `bird/`, `bird/[ring]/`, `pulli/`, `ring-sequences/`, `highlights/`) to accept an optional `viewedGroup?: { id: number; slug: string | null }` prop. **Do not destructure or consume it** in the function body — TypeScript's excess-property check on JSX requires the type to exist even though nothing reads it yet.

## Out of scope

- Renaming `app/group/[groupId]` → `app/group/[groupSlug]` (deferred to a later ticket not in this batch).
- Actually consuming `viewedGroup` anywhere (tickets 2 and 3).
[...]

Issue #472, opened 2026-08-18
Group slugs


wheresrhys:
## Group slugs in URLs

Replace the numeric `RingingGroups.id` in cross-group page URLs (`/group/<id>/...`) with a
human-readable, no-special-characters `slug`.

Full plan: `.claude/plans/i-woudl-like-group-parallel-brook.md` (local, not checked in).

### Sub-issues (implementation order)

1. Add nullable `slug` column + `slugify` util + seed/import-csv updates
2. Enforce `NOT NULL` on `slug` (blocked by 1, plus a manual human prod-backfill)
3. `lib/group-slug.ts` resolver + rename `app/group/[groupId]` → `app/group/[groupSlug]` (blocked by 1)
4. Thread slug through href-building components (blocked by 3)

PR #496, opened 2026-08-19
BootstrapPageData accepts viewedGroup input prop

### LINKED ISSUE #487: BootstrapPageData accepts viewedGroup input prop
## Context

Part of the "Group slugs in URLs" effort (#472), step 2 of 3 — depends on #486 (needs `lib/group-slug.ts` and the `viewedGroup` object already computed by the `/group` wrapper layer). Still does not change what `PageComponent` receives (that's ticket 3) — this ticket only changes `BootstrapPageData`'s *input* contract.

## Scope

- `app/components/layout/BootstrapPageData.tsx`: change `BootstrapPageDataProps`'s input from `viewedGroupId?: number` to `viewedGroup?: { id: number; slug: string | null }`.
- In `LoadWithData`: if `viewedGroup` isn't passed, derive one from the logged-in group cookie: `loggedInGroupId ? { id: loggedInGroupId, slug: await resolveGroupSlugById(loggedInGroupId) } : null`. Preserve the existing "Select a group to view data on this site" fallback when there's no logged-in group and no override.
- The internal numeric `viewedGroupId` local (used for cache keys and `dataFetcher`) derives from `viewedGroup.id` — `dataFetcher`'s own signature (`(params, viewedGroupId: number)`) is unchanged; only the wiring above it changes.
- `PageComponent`'s prop type stays `viewedGroupId: number` in this ticket — unchanged until ticket 3.
- Update `app/components/layout/__mocks__/BootstrapPageData.tsx` to accept `viewedGroup` in place of `viewedGroupId` on the input side (its `viewedGroupId={1}` call into `PageComponent` stays as-is for now — that's ticket 3).
- Update the 15 `app/(routes)/*/page.tsx` files from #486: each now *uses* the `viewedGroup` prop it already accepts — pass `viewedGroup={viewedGroup}` into its `<BootstrapPageData>` call instead of `viewedGroupId={viewedGroupId}`.

## Out of scope

- `PageComponent`'s own prop shape (ticket 3).
- The `/group` wrapper files' calls (still pass both `viewedGroup` and `viewedGroupId`, per #486 — ticket 3 drops the redundant one).
- Route rename.

## Files touched

- `app/components/layout/BootstrapPageData.tsx`, `app/components/layout/__mocks__/BootstrapPageData.tsx`
- 15 files under `app/(routes)/**`

## Commits

1. `BootstrapPageData`: swap `viewedGroupId` input for `viewedGroup`, derive default from the logged-in group cookie via `resolveGroupSlugById`; update mock
2. Update the 15 `app/(routes)/*/page.tsx` files to pass `viewedGroup` into `<BootstrapPageData>`

## Tests (USE)

- `BootstrapPageData`/`LoadWithData`:
  - Usual: `viewedGroup` passed explicitly → used as-is, no cookie-based resolution triggered.
  - Structure: `viewedGroup` omitted → derives `{ id: loggedInGroupId, slug }` from the cookie + `resolveGroupSlugById`.
  - Edge: no `viewedGroup` prop and no logged-in-group cookie → still renders the "Select a group" fallback (regression).
  - Edge: `viewedGroup` omitted, `loggedInGroupId` present but `resolveGroupSlugById` resolves `null` → `viewedGroup.slug` is `null`, no crash (defensive; shouldn't occur post-#480's `NOT NULL` slug, but the function's return type still allows it).
- Update existing `BootstrapPageData` tests that currently pass `viewedGroupId` to use `viewedGroup`.

wheresrhys:
## Summary

Part of the "Group slugs in URLs" effort (#472), step 2 of 3 — depends on #486
(`lib/group-slug.ts` + the `viewedGroup` object already computed by the `/group`
wrapper layer, merged via #495).

- `app/components/layout/BootstrapPageData.tsx`: `BootstrapPageDataProps`'s input
  changes from `viewedGroupId?: number` to `viewedGroup?: { id: number; slug: string | null }`.
  In `LoadWithData`, when `viewedGroup` isn't passed it's derived from the logged-in
  group cookie: `loggedInGroupId ? { id: loggedInGroupId, slug: await resolveGroupSlugById(loggedInGroupId) } : null`.
  The existing "Select a group to view data on this site" fallback is preserved.
  The internal numeric `viewedGroupId` (used for cache keys and `dataFetcher`) now
  derives from `viewedGroup.id` — `dataFetcher`'s own signature is unchanged.
- `PageComponent`'s prop type stays `viewedGroupId: number` in this ticket — unchanged
  until ticket 3 (#488).
- Updated the 15 `app/(routes)/*/page.tsx` files (from #486) to pass their
  already-accepted `viewedGroup` prop into `<BootstrapPageData>` instead of
  `viewedGroupId`.
- Also updated the two cross-group `/group/[groupId]/session/[date]{,/site/[locationId]}/page.tsx`
  wrapper pages, which call `BootstrapPageData` directly with `viewedGroupId` — these
  needed the same wiring change to keep compiling against the new contract (not one
  of the enumerated 15 files, but a direct consequence of the interface change).
- `app/components/layout/__mocks__/BootstrapPageData.tsx` needed no code change — it
  never destructures `viewedGroupId` on the input side, only inherits the (now updated)
  shared `BootstrapPageDataProps` type; its `viewedGroupId={1}` call into `PageComponent`
  stays as-is (that's ticket 3).

## Out of scope

- `PageComponent`'s own prop shape (ticket 3, #488).
- The `/group` wrapper files' route-page calls still pass both `viewedGroup` and
  `viewedGroupId` into the 15 route pages, per #486 — ticket 3 drops the redundant one.
- Route rename.

## Tests

`app/components/layout/__tests__/BootstrapPageData.test.tsx` (new) — exercises the real
`LoadWithData` (unmocked) via the USE cases from the ticket:
- Usual: explicit `viewedGroup` used as-is, no cookie-based resolution triggered.
- Structure: `viewedGroup` omitted → derives `{ id: loggedInGroupId, slug }` from the
  cookie + `resolveGroupSlugById`.
- Edge: no `viewedGroup` and no logged-in-group cookie → "Select a group" fallback
  (regression).
- Edge: `viewedGroup` omitted, `loggedInGroupId` present but `resolveGroupSlugById`
  resolves `null` → no crash.

`npm run qa` passes (lint, type-check, 738 app tests). Pre-push hook (91 E2E tests) passes.

Closes #487

### DIFF
--- a/app/(routes)/bird/[ring]/page.tsx
+++ b/app/(routes)/bird/[ring]/page.tsx
@@ -135,6 +135,7 @@ export default async function BirdPage(
 	return (
 		<BootstrapPageData<StandaloneBird, PageProps, PageParams>
 			pageProps={props}
+			viewedGroup={props.viewedGroup}
 			getParams={async (pageProps: PageProps) => ({
 				ring: (await pageProps.params).ring.toUpperCase()
 			})}
--- a/app/(routes)/bird/page.tsx
+++ b/app/(routes)/bird/page.tsx
@@ -35,6 +35,7 @@ export default async function BirdPage(
 	return (
 		<BootstrapPageData<SearchResult[], PageProps, SearchParams>
 			pageProps={props}
+			viewedGroup={props.viewedGroup}
 			getParams={async (pageProps: PageProps) => ({
 				q: (await pageProps.searchParams).q
 			})}
--- a/app/(routes)/controls/page.tsx
+++ b/app/(routes)/controls/page.tsx
@@ -17,14 +17,14 @@ async function dataFetcher(
 }
 
 export default async function ControlsRoute({
-	viewedGroupId
+	viewedGroup
 }: {
 	viewedGroupId?: number;
 	viewedGroup?: ViewedGroup;
 } = {}) {
 	return (
 		<BootstrapPageData<RingSequenceControlRow[]>
-			viewedGroupId={viewedGroupId}
+			viewedGroup={viewedGroup}
 			getCacheKeys={() => ['controls']}
 			dataFetcher={dataFetcher}
 			PageComponent={ControlsPage}
--- a/app/(routes)/effort/page.tsx
+++ b/app/(routes)/effort/page.tsx
@@ -128,14 +128,14 @@ function PayOffPageContent({
 }
 
 export default function PayOffPage({
-	viewedGroupId
+	viewedGroup
 }: {
 	viewedGroupId?: number;
 	viewedGroup?: ViewedGroup;
 } = {}) {
 	return (
 		<BootstrapPageData<PayOffStatsData>
-			viewedGroupId={viewedGroupId}
+			viewedGroup={viewedGroup}
 			getCacheKeys={() => ['sessions', 'pay-off']}
 			dataFetcher={fetchPayOffPageData}
 			PageComponent={PayOffPageContent}
--- a/app/(routes)/highlights/page.tsx
+++ b/app/(routes)/highlights/page.tsx
@@ -184,14 +184,14 @@ function HighlightsPageContent({
 }
 
 export default async function HighlightsPage({
-	viewedGroupId
+	viewedGroup
 }: {
 	viewedGroupId?: number;
 	viewedGroup?: ViewedGroup;
 } = {}) {
 	return (
 		<BootstrapPageData<StatsAccordionModel[]>
-			viewedGroupId={viewedGroupId}
+			viewedGroup={viewedGroup}
 			getCacheKeys={() => ['highlights']}
 			dataFetcher={fetchHighlightsData}
 			PageComponent={HighlightsPageContent}
--- a/app/(routes)/mistakes/page.tsx
+++ b/app/(routes)/mistakes/page.tsx
@@ -31,14 +31,14 @@ function ListMistakes({ data }: { data: DiscrepenciesResult[] }) {
 	);
 }
 export default async function MistakesPage({
-	viewedGroupId
+	viewedGroup
 }: {
 	viewedGroupId?: number;
 	viewedGroup?: ViewedGroup;
 } = {}) {
 	return (
 		<BootstrapPageData<DiscrepenciesResult[]>
-			viewedGroupId={viewedGroupId}
+			viewedGroup={viewedGroup}
 			getCacheKeys={() => ['mistakes']}
 			dataFetcher={fetchMistakes}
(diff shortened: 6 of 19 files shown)


wheresrhys:
## Behaviour change: how `LoadWithData` resolves "the group being viewed" when no override is passed

The prop rename (`viewedGroupId` → `viewedGroup`) on `BootstrapPageDataProps` looks
mechanical, but the fallback path it replaces isn't just a rename — it adds a real
async resolution step. Worked example: a route page renders with no `viewedGroup`
prop (the common case — most `app/(routes)/*` pages), and the visitor's
`selected_group_id` cookie points at group `7` ("Alpha Ringers", slug `alpha-ringers`).

**Before (#486 and earlier):** the fallback only ever needed the numeric id.

```mermaid
flowchart TB
    A["No viewedGroup/viewedGroupId prop passed"] --> B["getGroupCookie() → 7"]
    B --> C["viewedGroupId = 7"]
    C --> D["dataFetcher(params, 7)"]
    C --> E["PageComponent viewedGroupId=7"]
```

**After (this PR):** the same fallback now also resolves the slug, because
`BootstrapPageDataProps` no longer has a bare id — every path through `LoadWithData`
must produce a full `{ id, slug }` object.

```mermaid
flowchart TB
    A["No viewedGroup prop passed"] --> B["getGroupCookie() → 7"]
    B --> C{"loggedInGroupId set?"}
    C -->|"no"| F["viewedGroup = null"]
    F --> G["render: 'Select a group to view data on this site'"]
    C -->|"yes, 7"| D["resolveGroupSlugById(7) → 'alpha-ringers'"]
    D --> E["viewedGroup = { id: 7, slug: 'alpha-ringers' }"]
    E --> H["viewedGroupId = viewedGroup.id  (7)"]
    H --> I["dataFetcher(params, 7)  — signature unchanged"]
    H --> J["PageComponent viewedGroupId=7  — signature unchanged"]
```

Two things worth calling out:

- **Skipped when the caller passes `viewedGroup` explicitly** (the 15 `app/(routes)`
  pages once the `/group` wrapper layer supplies it, and the two cross-group
  `/group/.../session/...` pages) — `resolveGroupSlugById` is never called in that
  path, so there's no extra DB round-trip for the common logged-in-own-group case.
- **`dataFetcher`/`PageComponent` are untouched** — they still receive a plain numeric
  `viewedGroupId`, same as before. Only the *input* side of `BootstrapPageData` changed;
  this ticket deliberately stops short of threading the slug further down (that's #488).

PR #498, opened 2026-08-20
Thread viewedGroup into PageComponents; drop redundant viewedGroupId

### LINKED ISSUE #488: Thread viewedGroup into PageComponents; drop redundant viewedGroupId

wheresrhys:
## What this covers

Step 3 of 3 in the "Group slugs in URLs" effort (#472), building directly on #486 (`lib/group-slug.ts` resolver + `ViewedGroup` type) and #487 (`BootstrapPageData` accepts a `viewedGroup` input prop), both already merged.

This is where `viewedGroupId` disappears from the `PageComponent` boundary and from the `/group` wrapper layer's calls:

- **`BootstrapPageData`**: `PageComponent`'s prop type changes `viewedGroupId: number` → `viewedGroup: ViewedGroup`; `LoadWithData` now narrows on the resolved `viewedGroup` object directly (so the non-null object flows to the render) and passes `viewedGroup={viewedGroup}`. The `__mocks__` copy and its test render with `viewedGroup={{ id: 1, slug: 'alpha' }}`.
- **The 9 `PageComponent` implementations** (`HomePageContent`, `PayOffPageContent`, `ListAllSessions`, `BirdSummary`, `HighlightsPageContent`, `SpPage`, `SppStatsTable`, `ControlsPage`, `RingSequencesPage`) take `viewedGroup: ViewedGroup` and read `viewedGroup.id` where they previously used the plain id. Grandchildren (`SpStats`, `SessionHistoryCalendar`, `RingSizeSection`, …) keep receiving a plain `viewedGroupId: number` — the object shape does not cascade past the boundary.
- **The 10 `/group/[groupId]/**` wrapper files** drop the now-redundant `viewedGroupId={viewedGroupId}` prop, passing only `viewedGroup`.
- **The route-page override prop types** drop `viewedGroupId?: number` entirely — `viewedGroup` is the sole override input.

## Notes / assumptions (subagent mode)

- Used the shared `ViewedGroup` type (from `@/lib/group-slug`) for the new prop rather than re-spelling the inline `{ id: number; slug: string | null }` literal each time — it is exactly that shape, and the route pages already import it (DRY / consistency per CLAUDE.md). Client components import it type-only, so it is erased from the client bundle.
- `SessionSummary` (the session route's `PageComponent`) reads `viewedGroupId` from `params`, not a top-level prop, so it stays assignable to the boundary unchanged and needs no edit — consistent with the ticket excluding it.

## Tests

- `BootstrapPageData` test's `TestPageComponent` consumes `viewedGroup`; the two boundary-component tests (`SppStatsTable`, `RingSequencesPage`) render with `viewedGroup={{ id, slug }}`; the two `/group` wrapper tests drop the stale `viewedGroupId` assertion.
- Full app suite green (738 passing), `type:check` clean, lint 0 errors, E2E safe suite (91) green via the pre-push hook.

Closes #488.

### DIFF
--- a/app/(routes)/bird/[ring]/page.tsx
+++ b/app/(routes)/bird/[ring]/page.tsx
@@ -83,11 +83,10 @@ function BirdSummary({
 	data: StandaloneBird;
 	viewedGroup: ViewedGroup;
 }) {
-	const loggedInGroupId = viewedGroup.id;
 	const enrichedBird = bird.encounters.length ? enrichBird(bird) : null;
 	const sharedEncounterCount = enrichedBird
 		? enrichedBird.encounters.filter(
-				(e) => e.ringing_group_id !== loggedInGroupId
+				(e) => e.ringing_group_id !== viewedGroup.id
 			).length
 		: 0;
 	return (
--- a/app/(routes)/bird/[ring]/page.tsx
+++ b/app/(routes)/bird/[ring]/page.tsx
@@ -77,12 +77,13 @@ export async function fetchBirdData(
 function BirdSummary({
 	params: { ring },
 	data: bird,
-	viewedGroupId: loggedInGroupId
+	viewedGroup
 }: {
 	params: PageParams;
 	data: StandaloneBird;
-	viewedGroupId: number;
+	viewedGroup: ViewedGroup;
 }) {
+	const loggedInGroupId = viewedGroup.id;
 	const enrichedBird = bird.encounters.length ? enrichBird(bird) : null;
 	const sharedEncounterCount = enrichedBird
 		? enrichedBird.encounters.filter(
--- a/app/(routes)/controls/page.tsx
+++ b/app/(routes)/controls/page.tsx
@@ -19,7 +19,6 @@ async function dataFetcher(
 export default async function ControlsRoute({
 	viewedGroup
 }: {
-	viewedGroupId?: number;
 	viewedGroup?: ViewedGroup;
 } = {}) {
 	return (
--- a/app/(routes)/effort/page.tsx
+++ b/app/(routes)/effort/page.tsx
@@ -110,7 +110,7 @@ function PayOffPageContent({
 	data
 }: {
 	data: PayOffStatsData;
-	viewedGroupId: number;
+	viewedGroup: ViewedGroup;
 }) {
 	return (
 		<PageWrapper>
@@ -130,7 +130,6 @@ function PayOffPageContent({
 export default function PayOffPage({
 	viewedGroup
 }: {
-	viewedGroupId?: number;
 	viewedGroup?: ViewedGroup;
 } = {}) {
 	return (
--- a/app/(routes)/highlights/page.tsx
+++ b/app/(routes)/highlights/page.tsx
@@ -170,15 +170,15 @@ export async function fetchHighlightsData(
 
 function HighlightsPageContent({
 	data,
-	viewedGroupId
+	viewedGroup
 }: {
 	data: StatsAccordionModel[];
-	viewedGroupId: number;
+	viewedGroup: ViewedGroup;
 }) {
 	return (
 		<PageWrapper>
 			<PrimaryHeading>Highlights</PrimaryHeading>
-			<StatsAccordion data={data} viewedGroupId={viewedGroupId} />
+			<StatsAccordion data={data} viewedGroupId={viewedGroup.id} />
 		</PageWrapper>
 	);
 }
@@ -186,7 +186,6 @@ function HighlightsPageContent({
 export default async function HighlightsPage({
 	viewedGroup
 }: {
-	viewedGroupId?: number;
 	viewedGroup?: ViewedGroup;
 } = {}) {
 	return (
--- a/app/(routes)/mistakes/page.tsx
+++ b/app/(routes)/mistakes/page.tsx
@@ -33,7 +33,6 @@ function ListMistakes({ data }: { data: DiscrepenciesResult[] }) {
 export default async function MistakesPage({
 	viewedGroup
 }: {
-	viewedGroupId?: number;
 	viewedGroup?: ViewedGroup;
 } = {}) {
 	return (
--- a/app/(routes)/page.tsx
+++ b/app/(routes)/page.tsx
@@ -267,17 +267,17 @@ function TopSpecies({
 }
 function HomePageContent({
 	data,
-	viewedGroupId
+	viewedGroup
 }: {
 	data: PageModel;
-	viewedGroupId: number;
+	viewedGroup: ViewedGroup;
 }) {
 	return (
 		<PageWrapper>
 			<SummaryStatsTable data={data.summaryStats} />
 			<RecentSessions
 				data={data.recentSessions}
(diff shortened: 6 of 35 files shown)


wheresrhys:
## Behaviour change: `viewedGroup` object replaces `viewedGroupId` at the PageComponent boundary — and stops there

The value crossing the `BootstrapPageData → PageComponent` boundary changes shape from a bare number to a `{ id, slug }` object. The subtle part is **where the new shape stops**: it does *not* cascade — grandchildren below each PageComponent keep receiving a plain `viewedGroupId: number`, exactly as before. The `/group` wrapper layer also stops sending its duplicate `viewedGroupId` prop.

Worked example: viewing group `2` (slug `alpha`) on the species page.

```mermaid
flowchart TB
    subgraph OLD["Before"]
        direction TB
        w1["/group/[2] wrapper"] -->|"viewedGroupId=2<br/>viewedGroup={id:2,slug}"| r1["route page (SpeciesPage)"]
        r1 -->|"viewedGroup"| b1["BootstrapPageData"]
        b1 -->|"viewedGroupId=2"| p1["PageComponent (SpPage)"]
        p1 -->|"viewedGroupId=2"| g1["grandchildren<br/>(SpStats, tabs…)"]
    end
    subgraph NEW["After"]
        direction TB
        w2["/group/[2] wrapper"] -->|"viewedGroup={id:2,slug}<br/>(redundant id prop dropped)"| r2["route page (SpeciesPage)"]
        r2 -->|"viewedGroup"| b2["BootstrapPageData"]
        b2 -->|"viewedGroup={id:2,slug}"| p2["PageComponent (SpPage)"]
        p2 -->|"viewedGroupId = viewedGroup.id = 2"| g2["grandchildren<br/>(SpStats, tabs…)<br/>UNCHANGED: plain number"]
    end
    OLD --> NEW
```

What moved at each layer:

| Layer | Before | After |
|---|---|---|
| `/group/[groupId]/**` wrapper → route page | `viewedGroupId={n}` **and** `viewedGroup={…}` | `viewedGroup={…}` only |
| Route-page override prop type | `viewedGroupId?: number` + `viewedGroup?` | `viewedGroup?` only (sole override) |
| `BootstrapPageData` → `PageComponent` | `viewedGroupId: number` | `viewedGroup: ViewedGroup` |
| `PageComponent` → its children | `viewedGroupId: number` | `viewedGroupId: number` (unchanged — `viewedGroup.id` unwrapped at the boundary) |

Secondary effect inside `LoadWithData`: the empty-group guard now branches on the resolved `viewedGroup` object itself (`if (!viewedGroup)`) rather than on `viewedGroupId`, so the non-null object is what gets handed to the render.

app/(routes)/bird/[ring]/page.tsx
```diff
@@ -77,12 +77,13 @@ export async function fetchBirdData(
...
 	data: bird,
-	viewedGroupId: loggedInGroupId
+	viewedGroup
 }: {
 	params: PageParams;
 	data: StandaloneBird;
-	viewedGroupId: number;
+	viewedGroup: ViewedGroup;
 }) {
+	const loggedInGroupId = viewedGroup.id;
```
Comment (wheresrhys) [inline app/(routes)/bird/[ring]/page.tsx]: can we drop this and just use viewedGroup.id in the filter below

Comment (wheresrhys) [reply app/(routes)/bird/[ring]/page.tsx]: Done in bb434a6 — dropped the `loggedInGroupId` local and now use `viewedGroup.id` directly in the `sharedEncounterCount` filter. Bird page tests + full E2E suite pass.

Issue #490, opened 2026-08-19
Rename app/group/[groupId] to app/group/[groupSlug]; flip hrefs live


wheresrhys:
## Context

Part of the "Group slugs in URLs" effort (#472) — final ticket, depends on #489. Combines the route-rename half of the original #475 scope with flipping the two hrefs ticket 4 (#489) deliberately left pointing at `.id`. These two changes must land together: renaming the route alone would 404 every existing numeric-id link, and flipping the hrefs alone would generate slug-shaped links against a route that doesn't accept them yet.

## Scope

- `git mv` the 13 files under `app/group/[groupId]/**` to `app/group/[groupSlug]/**` (12 `page.tsx` + `session/_shared.tsx`).
- Each wrapper: parse `groupSlug` (string) from params instead of `groupId`; resolve `const viewedGroupId = await resolveGroupIdBySlug(groupSlug)`; if it resolves to `null` (no group has that slug), call `notFound()`.
- Build `viewedGroup = { id: viewedGroupId, slug: groupSlug }` directly from the resolved id and the URL's own slug (cheaper than #486's `resolveGroupSlugById` pattern, since the slug is already the param).
- Update the existing "redirect to `/` when viewing your own group" check to compare the resolved `viewedGroupId` against `loggedInGroupId`, same as today, just resolved via slug now.
- `app/components/shared/StatOutput.tsx`: flip the href from `${viewedGroup?.id}` to `${viewedGroup?.slug}`.
- `app/components/SessionHistoryCalendar.tsx`: flip the href from `${viewedGroup.id}` to `${viewedGroup.slug}`.

## Out of scope

- Anything already covered by #486–#489.

## Files touched

- 13 files, `git mv`'d and edited, under `app/group/[groupSlug]/**`
- `app/components/shared/StatOutput.tsx`, `app/components/SessionHistoryCalendar.tsx` (one-line href flips)

## Commits

1. Rename directory and update each wrapper to resolve by slug instead of id
2. Flip the two live hrefs from `.id` to `.slug`

## Tests (USE)

- Per wrapper (representative subset — plain `page.tsx` and one nested route):
  - Usual: valid `groupSlug` param → resolves to the correct numeric id, `viewedGroup` is `{ id, slug: groupSlug }`.
  - Edge: unknown `groupSlug` → `notFound()` triggers (404), not a crash.
  - Edge: `groupSlug` resolves to the logged-in group's own id → redirect-to-`/` regression, unchanged.
[...]

Issue #489, opened 2026-08-19
Cascade viewedGroup into the href-adjacent component tree (no behaviour change)


wheresrhys:
## Context

Part of the "Group slugs in URLs" effort (#472), between #488 (ticket 3) and the route-rename ticket that follows this one. #488 stopped threading `viewedGroup` at the `PageComponent` boundary; several components further downstream still take a plain `viewedGroupId` and eventually build an `href` with it. This ticket cascades the object one or two levels deeper into that specific chain — reusing #482's already-reviewed code — but deliberately keeps every `href` building from `viewedGroup.id` (unchanged output) rather than `viewedGroup.slug`. That keeps this ticket a pure no-op behaviourally and independently shippable to prod without breaking anything (the slug flip and the route rename land together in the following ticket, so the app is never mid-transition in production).

## Scope

Swap `viewedGroupId: number` for `viewedGroup: { id: number; slug: string | null }` in, and thread the object one level further through:

1. `app/components/SpPage.tsx` — `SpeciesData` passes `viewedGroup={viewedGroup}` to `SpStats` (its own top-level prop was already swapped in #488; this is the follow-on pass-through edit).
2. `app/components/SpStats.tsx` — prop swap; passes `viewedGroup={viewedGroup}` to `StatOutput`.
3. `app/components/shared/StatOutput.tsx` — prop swap; `href` keeps building `${viewedGroup?.id}` (same numeric output as today, just read from the object).
4. `app/components/StatsAccordion.tsx` — prop swap; `useEffect` dependency and the `key=` template use `viewedGroup.id`; passes `viewedGroup={viewedGroup}` to `StatsAccordionItem`.
5. `app/components/StatsAccordionItem.tsx` — prop swap; `ringing_group_filter: model.viewedGroup.id`; passes `viewedGroup={model.viewedGroup}` to `StatOutput`.
6. `app/components/SessionHistoryCalendar.tsx` — prop swap; `href` keeps building `${viewedGroup.id}`; passes `viewedGroup` to `MonthSessions`.
7. `app/components/MonthSessions.tsx` — prop swap; passes `viewedGroup` to `YearSessions`.
8. `app/components/YearSessions.tsx` — prop swap; passes `viewedGroup` through its own recursive month/year model.

## Out of scope

- Flipping any `href` to use `.slug` (next ticket).
- The route rename (next ticket).
- Any component not on this specific chain (already covered by #488, or never consumed `viewedGroupId`).

## Files touched

[...]

PR #502, opened 2026-08-20
Cascade viewedGroup into the href-adjacent component tree (no behaviour change)

wheresrhys:
## What this covers

Cascades the `viewedGroup` `{ id, slug }` object (introduced upstream in #486–#488) one or two levels deeper into the specific component chain that eventually builds an `href`, swapping each component's `viewedGroupId: number` prop for `viewedGroup: ViewedGroup`. Every `href`, accordion key, `useEffect` dependency and `ringing_group_filter` still reads `viewedGroup.id` — so the rendered output is **byte-identical** to before. This is a deliberate no-op behaviourally, independently shippable to prod; the slug flip and route rename land together in the following ticket (#490).

Implements #489 (child of #472).

### Components swapped (`viewedGroupId` → `viewedGroup`)
- `app/components/shared/StatOutput.tsx` — session `href` still `/group/${viewedGroup?.id}/...`
- `app/components/StatsAccordion.tsx` — `useEffect` dep + `key` use `viewedGroup.id`
- `app/components/StatsAccordionItem.tsx` — `ringing_group_filter: model.viewedGroup.id`
- `app/components/SessionHistoryCalendar.tsx` (incl. `SessionsByDay`) — `href` still `/group/${viewedGroup.id}/...`
- `app/components/YearSessions.tsx` — threads the object through the recursive year/month accordion models
- `app/components/MonthSessions.tsx`
- `app/components/SpStats.tsx`
- `app/components/SpPage.tsx` (`SpeciesData`) — passes the object to `SpStats`, the plain `viewedGroup.id` to the other species tabs (unchanged)

### Caller edits (necessary consequence of the prop swaps)
The pages/components that render the swapped components now pass the object they already hold:
- `app/(routes)/highlights/page.tsx` → `StatsAccordion`
- `app/(routes)/sessions/page.tsx` → `SessionHistoryCalendar`
- `app/(routes)/page.tsx` (`RecentSessions`) → `SessionsByDay` (home page also consumes `SessionsByDay`)

## Tests

- Updated existing component tests to render with `viewedGroup={{ id, slug: 'alpha' }}` instead of `viewedGroupId={id}` (`SpStats`, `StatsAccordionItem`, `YearSessions`).
- Added `app/components/shared/__tests__/StatOutput.test.tsx` — byte-exact href assertions (`/group/1/session/2022-10-20`, plus the `/site/31` variant) and the missing-`viewedGroup` throw.
- Added a `YearSessions` regression test asserting the calendar's threaded session href stays `/group/1/session/...` and no href leaks the slug.
- `npm run qa` (lint + type-check + 742 app tests) green; pre-push E2E suite (91 tests) green.

- `SessionsByDay` (living in `SessionHistoryCalendar.tsx`) is the actual href-builder the ticket's item 6 refers to, so it was swapped too; this necessarily threaded `viewedGroup` into its second consumer, the home page's `RecentSessions`. The home page's `BootstrapPageData` boundary (from #488) already supplies the object, so no behaviour changed.
- The ticket's "Files touched" list under-counted these two caller pages; they are unavoidable given the prop swaps and were kept minimal (passing the already-available object).

### DIFF
--- a/app/(routes)/highlights/page.tsx
+++ b/app/(routes)/highlights/page.tsx
@@ -178,7 +178,7 @@ function HighlightsPageContent({
 	return (
 		<PageWrapper>
 			<PrimaryHeading>Highlights</PrimaryHeading>
-			<StatsAccordion data={data} viewedGroupId={viewedGroup.id} />
+			<StatsAccordion data={data} viewedGroup={viewedGroup} />
 		</PageWrapper>
 	);
 }
--- a/app/(routes)/page.tsx
+++ b/app/(routes)/page.tsx
@@ -133,10 +133,10 @@ export async function fetchHomePageData(
 
 function RecentSessions({
 	data,
-	viewedGroupId
+	viewedGroup
 }: {
 	data: SessionWithEncountersCount[];
-	viewedGroupId: number;
+	viewedGroup: ViewedGroup;
 }) {
 	return (
 		<div>
@@ -152,7 +152,7 @@ function RecentSessions({
 			<BoxyList>
 				<SessionsByDay
 					sessions={data}
-					viewedGroupId={viewedGroupId}
+					viewedGroup={viewedGroup}
 					dateFormat="EEEE do MMMM"
 				/>
 			</BoxyList>
@@ -275,10 +275,7 @@ function HomePageContent({
 	return (
 		<PageWrapper>
 			<SummaryStatsTable data={data.summaryStats} />
-			<RecentSessions
-				data={data.recentSessions}
-				viewedGroupId={viewedGroup.id}
-			/>
+			<RecentSessions data={data.recentSessions} viewedGroup={viewedGroup} />
 			<TopSpecies data={data.topSpecies} lastGroupTick={data.lastGroupTick} />
 		</PageWrapper>
 	);
--- a/app/(routes)/sessions/page.tsx
+++ b/app/(routes)/sessions/page.tsx
@@ -38,7 +38,7 @@ function ListAllSessions({
 	return (
 		<PageWrapper>
 			<PrimaryHeading>Session history</PrimaryHeading>
-			<SessionHistoryCalendar sessions={data} viewedGroupId={viewedGroup.id} />
+			<SessionHistoryCalendar sessions={data} viewedGroup={viewedGroup} />
 		</PageWrapper>
 	);
 }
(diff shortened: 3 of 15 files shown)


wheresrhys:
## Behaviour change: none — that's the point

This PR swaps a `viewedGroupId: number` prop for a `viewedGroup: { id, slug }` object across 8 components (and 3 caller edits), but every URL it produces is **byte-identical**. The one thing worth seeing is *where the object now flows* vs *what it's allowed to read* — the deliberate boundary the next ticket (#490) will cross.

### 1. The object cascades one/two levels deeper; the href still reads `.id`

The prop shape changed the whole way down each chain, but the only field consumed at the leaf is still the numeric `id`. `slug` rides along, unread.

```mermaid
flowchart TB
  subgraph chainA["species chain"]
    A1["SpPage\nviewedGroup {id,slug}"] --> A2["SpStats\nviewedGroup"]
    A2 --> A3["StatOutput"]
  end
  subgraph chainB["highlights chain"]
    B1["StatsAccordion\nviewedGroup"] --> B2["StatsAccordionItem\nviewedGroup"]
    B2 --> B3["StatOutput"]
  end
  subgraph chainC["session calendar chain"]
    C1["SessionHistoryCalendar\nviewedGroup"] --> C2["YearSessions\nviewedGroup"]
    C2 --> C3["MonthSessions\nviewedGroup"]
    C3 --> C4["SessionsByDay\nviewedGroup"]
    C4 --> C5["StatOutput"]
  end
  A3 --> HREF["href = /group/{viewedGroup.id}/session/...\n(reads .id only — .slug ignored)"]
  B3 --> HREF
  C5 --> HREF
```

### 2. Same URL in, same URL out

Concrete leaf example, before and after — the object is threaded, the render is unchanged.

```mermaid
flowchart TB
  subgraph before["BEFORE"]
    OB["viewedGroupId = 1"] --> OBH["/group/1/session/2022-10-20"]
  end
  subgraph after["AFTER (this PR)"]
    OA["viewedGroup = { id: 1, slug: 'alpha' }"] --> OAH["/group/1/session/2022-10-20"]
  end
  subgraph next["NEXT TICKET #490 (not here)"]
    ON["viewedGroup = { id: 1, slug: 'alpha' }"] --> ONH["/group/alpha/session/2022-10-20"]
  end
```

| | reads | href |
|---|---|---|
| before | `viewedGroupId` | `/group/1/session/...` |
| **this PR** | `viewedGroup.id` | `/group/1/session/...` (identical) |
| next ticket #490 | `viewedGroup.slug` | `/group/alpha/session/...` |

The regression tests (`StatOutput`, `YearSessions`) pin the middle row: href stays `/group/1/...` and **no** href leaks `/group/alpha/...`, proving the 8-file swap is a true no-op.

PR #504, opened 2026-08-20
Rename app/group/[groupId] to app/group/[groupSlug]; flip hrefs live

### LINKED ISSUE #490: Rename app/group/[groupId] to app/group/[groupSlug]; flip hrefs live

wheresrhys:
## Summary

Final ticket of the "Group slugs in URLs" effort (#472) — renames the `app/group/[groupId]`
route tree to `app/group/[groupSlug]` and flips the last two live hrefs that #489 deliberately
left pointing at `.id`, landing both together since renaming the route alone would 404 every
existing link, and flipping the hrefs alone would generate slug-shaped links against a route
that didn't accept them yet.

- `git mv`'d the 13 files under `app/group/[groupId]/**` to `app/group/[groupSlug]/**`. Each
  wrapper now parses `groupSlug` from params, resolves it to the numeric id via
  `resolveGroupIdBySlug`, calls `notFound()` if no group has that slug, and builds
  `viewedGroup = { id, slug: groupSlug }` directly from the URL's own slug. The home wrapper's
  existing "redirect to `/` when viewing your own group" check now compares against the
  resolved id.
- Flipped `app/components/shared/StatOutput.tsx` and `app/components/SessionHistoryCalendar.tsx`
  hrefs from `viewedGroup.id` to `viewedGroup.slug`, and updated their href-assertion tests
  (`StatOutput.test.tsx`, `YearSessions.test.tsx`).
- Updated the two e2e specs that hardcoded numeric ids in `/group/<id>/...` URLs
  (`cross-group.spec.ts`, `cross-group-pages.spec.ts`) to use each seed group's slug instead, via
  a new `e2e/helpers/group-slugs.ts` sourced from the DB at global setup (mirroring the existing
  `group-ids.ts` pattern), rather than assuming what `slugify()` would produce.

## Known follow-up gap (assumption, flagged for review)

`app/group/[groupSlug]/session/_shared.tsx`'s own `Locations` and `SessionNavigation` link
builders (used only by the session detail page's "Previous/Next" and per-location badges) still
build hrefs from the numeric `viewedGroupId`, unchanged. This is **explicitly out of #490's
stated scope** — the ticket's "Files touched" list names only `StatOutput.tsx` and
`SessionHistoryCalendar.tsx` for href flips, and the ticket's own Tests section (mirrored in the
pre-existing `session/[date]/__tests__/page.test.tsx`) still asserts those two links use the
numeric id, not the slug. I left this test assertion and the underlying behaviour unchanged
rather than expand scope, but flagging: after this rename, in-session navigation
(Previous/Next session links, per-location badges when a session has multiple locations) will
generate `/group/<numeric-id>/...` URLs that 404 against the renamed route, until a follow-up
ticket flips those two remaining hrefs too.

## Test plan

- [x] `npm run type:check` — passes
- [x] `npm run test:nowatch` — 749 tests passed (64 files)
- [x] `npm run lint` — 0 errors (10 pre-existing warnings, unrelated to this change)
- [x] `npm run test:e2e:safe` (via pre-push hook's diff-aware selection) — 91 passed, including
      the updated `cross-group.spec.ts` / `cross-group-pages.spec.ts` now navigating by slug

Closes #490


wheresrhys:
## Behaviour change: `/group/[groupId]` → `/group/[groupSlug]`

### 1. The wrapper now resolves slug → id, not id → slug — and resolution moves *before* the redirect check

Before, the home wrapper parsed the numeric id synchronously, compared it to the logged-in
group straight away, and only resolved a slug afterwards (purely for display, on the
"viewing someone else" branch). After, the id itself has to come from an async DB lookup on the
slug, so that lookup now has to happen *before* the redirect can be evaluated at all — and it
also gains a new outcome the old code never had: an unknown slug.

```mermaid
flowchart TB
    subgraph before["Before — /group/2 (numeric)"]
        B1["viewedGroupId = Number('2')"] --> B2{"== loggedInGroupId?"}
        B2 -->|yes, e.g. own id| B3["redirect('/')"]
        B2 -->|no| B4["resolveGroupSlugById(2)<br/>→ 'beta' (for display only)"]
        B4 --> B5["render Home"]
    end

    subgraph after["After — /group/beta (slug)"]
        A1["resolveGroupIdBySlug('beta')"] --> A2{"found?"}
        A2 -->|null, unknown slug| A3["notFound() — 404"]
        A2 -->|resolved, e.g. 2| A4{"== loggedInGroupId?"}
        A4 -->|yes| A5["redirect('/')"]
        A4 -->|no| A6["viewedGroup = { id: 2, slug: 'beta' }<br/>render Home"]
    end
```

Every one of the 8 simple wrappers (effort, mistakes, pulli, resightings, retraps, ticks,
sessions, species) and the two session-detail wrappers follow the same shape as the right-hand
side, minus the redirect branch (only the home wrapper has that check).

### 2. The two hrefs #489 deliberately left as `.id` now emit slugs

```mermaid
flowchart TB
    subgraph before2["Before"]
        C1["viewedGroup = { id: 2, slug: 'beta' }"] --> C2["href = /group/2/session/2024-05-10"]
    end
    subgraph after2["After"]
        D1["viewedGroup = { id: 2, slug: 'beta' }"] --> D2["href = /group/beta/session/2024-05-10"]
    end
```

This is `StatOutput.tsx`'s single-session-day link and `SessionHistoryCalendar.tsx`'s
multi-session-day location link — the only two places in the live app that still built a
`/group/...` href from `viewedGroup.id` after #489.

### 3. Known gap this PR intentionally leaves: session-page internal navigation still 404s

`session/_shared.tsx`'s `Locations` badges and `SessionNavigation` Previous/Next links were
**not** in #490's scope (not listed in its "Files touched", and its own Tests section still
asserts the numeric id there) — but they build hrefs the same way the two flipped components
used to, so after this rename they now point at a route that no longer accepts numeric ids:

```mermaid
flowchart TB
    E1["On /group/beta/session/2024-05-10<br/>(rendered via resolved viewedGroupId=2)"] --> E2["'Previous' link<br/>href = /group/2/session/2024-04-20"]
    E2 --> E3["/group/[groupSlug] wrapper:<br/>resolveGroupIdBySlug('2') → null"]
    E3 --> E4["notFound() — 404"]
```

Flagged in the PR body as a follow-up; not fixed here to stay within #490's stated scope.

app/components/shared/StatOutput.tsx
```diff
@@ -57,7 +57,7 @@ export function StatOutput({
 			{temporalUnit === 'day' && link ? (
 				<NoPrefetchLink
 					className="link"
-					href={`/group/${viewedGroup?.id}/session/${visitDate}${location ? `/site/${location.id}` : ''}`}
+					href={`/group/${viewedGroup?.slug}/session/${visitDate}${location ? `/site/${location.id}` : ''}`}
```
wheresrhys, line 60: No need for ?. here I think

Issue #509, opened 2026-08-20
Add fetchYearStats and fetchMonthStats to lib/underlying-stats.ts


wheresrhys:
## Context

Follow-up to #507 (F1). Original spec asked for a generic `fetchPeriodStats(from, to)`; superseded by this design because summing per-day encounter/bird counts across an arbitrary range double-counts retrapped birds. The existing `aggregate_stats` RPC (`supabase/schema/schemas/public/functions/aggregate_stats.sql`) already avoids this — its `bird_count` is `COUNT(DISTINCT bird_id)` computed per period bucket server-side, not summed client-side. `app/actions/pay-off-stats.ts`'s `fetchPayOffStats` already calls this RPC exactly this way (`group_by_time_period: 'year'|'month'`, no `from_date`/`to_date` — the RPC generates one row per period with data automatically via its internal `period_spine`). This ticket adds the equivalent pair as reusable `lib/underlying-stats.ts` exports.

No schema/migration needed — `aggregate_stats` is unchanged.

## Scope

- Add to `lib/underlying-stats.ts`:
  ```ts
  export async function fetchYearStats(viewedGroupId: number): Promise<AggregateStatsResult[] | null> {
    const supabase = await getAuthenticatedSupabaseClient();
    return supabase.rpc('aggregate_stats', {
      ringing_group_filter: viewedGroupId,
      group_by_species: false,
      group_by_time_period: 'year'
    }).then(catchSupabaseErrors) as Promise<AggregateStatsResult[] | null>;
  }
  ```
  and `fetchMonthStats` identically with `group_by_time_period: 'month'`.
- `group_by_species: false` matches `fetchPayOffStats`'s existing precedent (one row per period, not per period+species) — reuse that convention rather than inventing a per-species variant, since nothing in this batch needs species breakdown.
- No caching (unlike `fetchSessionStats`) — `fetchPayOffStats` doesn't cache either; matching that precedent for consistency. This is a deliberate choice, not an oversight.
- Import `AggregateStatsResult` type from `@/app/models/db`.
- **Note for implementers of other tickets in this batch**: nothing else in this round (B/D/E) actually consumes `fetchYearStats`/`fetchMonthStats` — they only need presence/absence of session dates (from F1) or session date lists, not aggregate counts. This ticket exists to satisfy the original spec's ask for period-level stats functions and to get the retrap-safe pattern in place for future use.

## Out of scope

Any RPC/schema change; `group_by_species: true` variants; wiring into any page.

[...]

PR #515, opened 2026-08-20
feat: add lib/underlying-stats.ts with a copy of fetchSessionStats

### LINKED ISSUE #507: Add lib/underlying-stats.ts with a copy of fetchSessionStats
## Context

First of two lib tickets (F1/F2; tickets B/D/E build on this file later). `app/actions/session-highlights.ts` has a private `fetchSessionStats` (all-time per-day-per-species stats + session dates, cached per group). `session-highlights.ts` is slated for eventual deletion as part of a longer-term rework, so the new shared module must not import from it or share its cache — this is an intentional literal duplicate, not a refactor. Do not "fix" this into a shared import during this ticket.

## Scope

- Create `lib/underlying-stats.ts` (plain shared module, no `'use server'` — matches `lib/group-slug.ts`'s precedent of calling `getAuthenticatedSupabaseClient()` with no directive; `session-highlights.ts` has `'use server'` only because every export there is a server action, which isn't the case here).
- Copy `fetchStatsVersion`, `sessionStatsCache` (module-level `Map`), `SESSION_STATS_CACHE_TTL_MS`, and `fetchSessionStats(viewedGroupId: number): Promise<SessionStatsData>` from `app/actions/session-highlights.ts`, unchanged logic, own cache instance, exported.
- Import `SessionStatsData` type from `@/app/models/session-highlights` (type-only import across the `app/models` → `lib` boundary).
- Tests in `lib/__tests__/underlying-stats.test.ts`, matching `lib/__tests__/group-slug.test.ts`'s mocking pattern.

## Out of scope

- Changing `app/actions/session-highlights.ts` itself — it keeps its own private `fetchSessionStats` and cache, untouched.
- `fetchYearStats`/`fetchMonthStats` (ticket F2, depends on this one).
- Wiring this into any page or action (tickets B/D/E, which depend on this one).
- Deduplicating the two `fetchSessionStats` copies — intentional temporary duplication, see Context.

## Stack layers touched

`lib/` only — new shared module. No `app/actions/`, `app/components/`, route, or DB schema changes.

## Commit breakdown

1. `lib/underlying-stats.ts`: copy `fetchStatsVersion` + `fetchSessionStats` + cache map from `app/actions/session-highlights.ts`, unchanged logic, own cache instance.
2. `lib/__tests__/underlying-stats.test.ts` covering the export (USE enumeration below).

## Tests (USE algorithm)

```
describe('fetchSessionStats')
  it('fetches day-species stats and session dates for a group on a cold cache')
  it('returns the cached result on a second call when the stats version is unchanged')
  it('re-fetches when the stats version (max Encounters.id) has advanced since the cached entry')
  it('re-fetches when the cached entry has passed its TTL even if the version is unchanged')
  it('paginates through fetchAllPaginatedRows results beyond a single page')
  it('keeps separate cache entries per viewedGroupId')
```

## Reuse

`fetchAllPaginatedRows`, `catchSupabaseErrors` (`lib/supabase.ts`), `getAuthenticatedSupabaseClient` (`lib/group-auth.ts`) — all reused as-is, no changes needed.

wheresrhys:
## Why

Tickets B/D/E and the year/month-stats follow-up (#509) need a shared, non-server-action home for the day-species stats fetch that currently lives only as a private helper inside `app/actions/session-highlights.ts`. That module is slated for eventual deletion as part of a longer-term rework, so new code must not depend on it.

## What

Adds `lib/underlying-stats.ts`, a plain module (no `'use server'`, matching `lib/group-slug.ts`'s precedent) exporting `fetchStatsVersion`, `sessionStatsCache`, `SESSION_STATS_CACHE_TTL_MS`, and `fetchSessionStats(viewedGroupId)`. `app/actions/session-highlights.ts` now imports `fetchSessionStats` from this module instead of keeping its own private copy, so there's a single implementation and a single shared cache instance.

## Details

- `SessionStatsData` is imported as a type-only import from `@/app/models/session-highlights` across the `app/models` → `lib` boundary, per the ticket.
- Internal `lib`→`lib` imports (`./group-auth`, `./supabase`) use relative paths, not the `@/` alias — `tsconfig.json`'s `include` list doesn't cover `lib/**`, so `vite-tsconfig-paths` can't resolve `@/lib/...` specifiers from within `lib/`; existing `lib/group-slug.ts` already follows this relative-import convention for the same reason.
- `app/actions/session-highlights.ts` is `'use server'`; importing a plain (non-`'use server'`) module from it is fine, so it imports `fetchSessionStats` directly with no re-export shim.
- Tests in `lib/__tests__/underlying-stats.test.ts` mirror `lib/__tests__/group-slug.test.ts`'s mocking pattern, covering: cold-cache fetch, cache hit when version unchanged, re-fetch on version bump, re-fetch on TTL expiry, pagination beyond one page, and per-group cache isolation. `app/actions/__tests__/session-highlights.test.ts`'s existing caching-behaviour tests continue to pass unmodified, now exercising the shared implementation through the import rather than a private copy.

## Out of scope

- `fetchYearStats`/`fetchMonthStats` (ticket F2 / #509, depends on this one).
- Wiring this into any other page or action beyond `session-highlights.ts` (tickets B/D/E).

Closes #507

### DIFF
--- a/app/actions/session-highlights.ts
+++ b/app/actions/session-highlights.ts
@@ -1,6 +1,7 @@
 'use server';
 import { getAuthenticatedSupabaseClient } from '@/lib/group-auth';
-import { catchSupabaseErrors, fetchAllPaginatedRows } from '@/lib/supabase';
+import { catchSupabaseErrors } from '@/lib/supabase';
+import { fetchSessionStats } from '@/lib/underlying-stats';
 import { runHighlightMachine } from '@/app/models/highlight-refinement-machine';
 import {
 	deriveFirstEverSpecies,
@@ -13,82 +14,9 @@ import {
 	deriveSpeciesRecords,
 	deriveSpeciesJuvRecords,
 	deriveWeightRecordBreakers,
-	type SessionHighlight,
-	type SessionStatsData
+	type SessionHighlight
 } from '@/app/models/session-highlights';
-import type {
-	LongAbsenceRetrapsResult,
-	StatsPerDayAndSpeciesResult
-} from '@/app/models/db';
-
-// The stats blob only changes when new data is imported. The cache entry
-// carries a version token (max Encounters.id for the group) so that a new
-// import invalidates the cache immediately, even across lambda instances.
-// The 60-min TTL is retained as a backstop for rare in-place re-imports
-// that edit existing encounters without adding rows.
-const SESSION_STATS_CACHE_TTL_MS = 60 * 60 * 1000;
-const sessionStatsCache = new Map<
-	number,
-	{ version: number; expiresAt: number; stats: SessionStatsData }
->();
-
-async function fetchStatsVersion(
-	supabase: Awaited<ReturnType<typeof getAuthenticatedSupabaseClient>>,
-	viewedGroupId: number
-): Promise<number> {
-	const { data } = await supabase
-		.from('Encounters')
-		.select('id')
-		.eq('ringing_group_id', viewedGroupId)
-		.order('id', { ascending: false })
-		.limit(1);
-	return data?.[0]?.id ?? 0;
-}
-
-async function fetchSessionStats(
-	viewedGroupId: number
-): Promise<SessionStatsData> {
-	const supabase = await getAuthenticatedSupabaseClient();
-	const currentVersion = await fetchStatsVersion(supabase, viewedGroupId);
-	const cached = sessionStatsCache.get(viewedGroupId);
-	if (
-		cached &&
-		cached.version === currentVersion &&
-		cached.expiresAt > Date.now()
-	) {
-		return cached.stats;
-	}
-	const [daySpeciesStats, sessionRows] = await Promise.all([
-		fetchAllPaginatedRows<StatsPerDayAndSpeciesResult>((fromRow, toRow) =>
-			supabase
-				.rpc('stats_per_day_and_species', {
-					ringing_group_filter: viewedGroupId
-				})
-				.order('visit_date')
-				.order('species_name')
-				.range(fromRow, toRow)
-		),
-		fetchAllPaginatedRows<{ visit_date: string }>((fromRow, toRow) =>
-			supabase
-				.from('Sessions')
-				.select('visit_date')
-				.eq('ringing_group_id', viewedGroupId)
-				.eq('session_type', 'FULL_GROWN')
-				.order('visit_date')
-				.range(fromRow, toRow)
-		)
-	]);
-	const stats: SessionStatsData = {
-		daySpeciesStats,
-		sessionDates: sessionRows.map((row) => row.visit_date)
-	};
-	sessionStatsCache.set(viewedGroupId, {
-		version: currentVersion,
-		expiresAt: Date.now() + SESSION_STATS_CACHE_TTL_MS,
-		stats
-	});
-	return stats;
-}
+import type { LongAbsenceRetrapsResult } from '@/app/models/db';
 
 export async function fetchSessionHighlights({
 	date,
--- a/lib/__tests__/underlying-stats.test.ts
+++ b/lib/__tests__/underlying-stats.test.ts
@@ -1,0 +1,211 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import type { StatsPerDayAndSpeciesResult } from '@/app/models/db';
+
+const { mockGetAuthenticatedSupabaseClient } = vi.hoisted(() => ({
+	mockGetAuthenticatedSupabaseClient: vi.fn()
+}));
+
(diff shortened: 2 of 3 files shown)


wheresrhys:
## Behaviour change: a second, independent stats cache now exists

This PR doesn't wire anything into `lib/underlying-stats.ts` yet (that's tickets B/D/E and #509), but it's worth being explicit about what "intentional literal duplicate" means at runtime once those tickets land: `fetchSessionStats` for the same group can now be served from **two separate module-scoped caches** that never see each other's writes.

**Before this PR** — one cache, one source of truth for a group's day-species stats:

```mermaid
flowchart TB
    subgraph before["Today"]
        A["fetchSessionHighlights(groupId)"] --> B["app/actions/session-highlights.ts<br/>private sessionStatsCache"]
        B --> C[("stats_per_day_and_species RPC")]
    end
```

**After this PR** — a second caller path (future tickets) hits a distinct cache instance, so the same group's stats can be fetched and cached twice in parallel:

```mermaid
flowchart TB
    subgraph after["Once B/D/E and #509 wire this in"]
        D["fetchSessionHighlights(groupId)"] --> E["app/actions/session-highlights.ts<br/>private sessionStatsCache"]
        F["fetchYearStats / fetchMonthStats<br/>(future, #509)"] --> G["lib/underlying-stats.ts<br/>sessionStatsCache (new instance)"]
        E --> H[("stats_per_day_and_species RPC")]
        G --> H
    end
```

Both caches use the identical version+TTL invalidation logic (max `Encounters.id` per group, 60-min backstop), so correctness is unaffected — but a group viewed through both code paths pays for two RPC round-trips and holds two copies of the same blob in memory until the caches are eventually consolidated (out of scope here, see the ticket's "Out of scope" section).

No existing behaviour changes: `app/actions/session-highlights.ts` and everything that calls it are untouched.

wheresrhys:
Please can you also edit session-highlights.ts to simply import this function so that we get back to a single implementation

wheresrhys:
## Behaviour change: back to a single stats cache (supersedes the two-cache comment above)

The [earlier comment](#issuecomment-5358578609) flagged a risk from the *planned* two-cache future once tickets B/D/E and #509 wired callers into `lib/underlying-stats.ts`: two independent `sessionStatsCache` instances serving the same group's stats, never seeing each other's writes.

That risk no longer applies. This PR now has `app/actions/session-highlights.ts` import `fetchSessionStats` (and the `sessionStatsCache`/`fetchStatsVersion`/`SESSION_STATS_CACHE_TTL_MS` it depends on) directly from `lib/underlying-stats.ts`, instead of keeping its own private copy. There is one implementation and one cache instance, shared by every caller — today and future ones alike:

```mermaid
flowchart TB
    subgraph consolidated["Now — single shared cache"]
        A["fetchSessionHighlights(groupId)<br/>app/actions/session-highlights.ts"] --> S["fetchSessionStats(groupId)<br/>lib/underlying-stats.ts<br/>single sessionStatsCache"]
        F["fetchYearStats / fetchMonthStats<br/>(future, #509)"] -.-> S
        S --> C[("stats_per_day_and_species RPC")]
    end
```

`session-highlights.ts` keeps its `'use server'` directive — it just imports a plain module (`lib/underlying-stats.ts` has no `'use server'` of its own), which is fine since server actions can freely import plain helper modules. A group's stats are now fetched and cached exactly once no matter how many code paths call `fetchSessionStats`, so the double RPC round-trip / duplicated in-memory blob described above can't happen.

wheresrhys:
Done — `app/actions/session-highlights.ts` now imports `fetchSessionStats` (and the `sessionStatsCache` / `fetchStatsVersion` / `SESSION_STATS_CACHE_TTL_MS` it relies on) from `lib/underlying-stats.ts` instead of keeping its own private copy. There's a single implementation and a single shared cache instance now, rather than the incremental two-cache plan.

`session-highlights.ts` keeps its `'use server'` directive at the top — importing a plain (non-`'use server'`) module from it is fine, so no re-export shim was needed. `lib/underlying-stats.ts` itself stays a plain module, unchanged apart from a comment update.

Tests: `app/actions/__tests__/session-highlights.test.ts`'s existing caching-behaviour tests (cache hit on unchanged version, re-fetch on version bump, version query per call, etc.) needed no changes — they're black-box against `fetchSessionHighlights` and continue to pass, now exercising the shared implementation through the import rather than a private copy. `lib/__tests__/underlying-stats.test.ts` is unchanged and remains the source of truth for `fetchSessionStats`'s own behaviour. Ran `npm run qa` (lint + type-check + all app tests) clean — 68 files / 782 tests passing.

Also updated the PR description's "What"/"Details"/"Out of scope" sections to describe the consolidated single-cache state instead of the two-copy plan, and posted a follow-up diagram superseding the earlier two-cache-risk comment.

PR #521, opened 2026-08-20
feat: add fetchYearStats and fetchMonthStats to lib/underlying-stats.ts

### LINKED ISSUE #509: Add fetchYearStats and fetchMonthStats to lib/underlying-stats.ts

wheresrhys:
## Why

Follow-up to #507 (F1). The original spec asked for a generic `fetchPeriodStats(from, to)`, but summing per-day encounter/bird counts across an arbitrary range would double-count retrapped birds. The existing `aggregate_stats` RPC already avoids this by computing `bird_count` as `COUNT(DISTINCT bird_id)` per period bucket server-side — `fetchPayOffStats` (`app/actions/pay-off-stats.ts`) already calls it this way. This PR adds the equivalent pair as reusable `lib/underlying-stats.ts` exports for future callers.

## What

Adds `fetchYearStats` and `fetchMonthStats` to `lib/underlying-stats.ts`, calling `aggregate_stats` with `group_by_species: false` and `group_by_time_period: 'year'|'month'` respectively (no `from_date`/`to_date`, matching `fetchPayOffStats`' existing precedent exactly). No caching, matching `fetchPayOffStats` (a deliberate choice, not an oversight). No schema/migration/RPC changes.

### Assumption

The ticket's test spec titled one case per function "returns null when the RPC call errors," but the shared `catchSupabaseErrors` helper (`lib/supabase.ts`, unchanged/pre-existing) always throws on a Supabase error rather than returning null. Implemented the code exactly as specified in the ticket (mirroring `fetchPayOffStats`), and wrote those two tests to assert the actual (rejection) behaviour instead of the ticket's literal wording.

## Tests

Added to `lib/__tests__/underlying-stats.test.ts` (same file as F1's `fetchSessionStats` tests, new describe blocks): RPC args (time period, `group_by_species: false`), `ringing_group_filter` scoping, no `from_date`/`to_date`, and error-throw behaviour, for both `fetchYearStats` and `fetchMonthStats`.

`npm run qa` (lint + type-check + app tests) passes: 798 tests / 68 files.

Closes #509

### DIFF
--- a/lib/underlying-stats.ts
+++ b/lib/underlying-stats.ts
@@ -6,7 +6,7 @@ import type {
 	StatsPerDayAndSpeciesResult
 } from '@/app/models/db';
 
-// The stats blob only changes when new data is imported. The cache entry
+// The stats blobs only change when new data is imported. Each cache entry
 // carries a version token (max Encounters.id for the group) so that a new
 // import invalidates the cache immediately, even across lambda instances.
 // The 60-min TTL is retained as a backstop for rare in-place re-imports
@@ -16,10 +16,28 @@ import type {
 // session-highlights.ts imports it (and shares this cache instance) rather
 // than keeping its own copy, since importing a plain module from a
 // 'use server' file is fine.
-export const SESSION_STATS_CACHE_TTL_MS = 60 * 60 * 1000;
+//
+// fetchYearStats/fetchMonthStats reuse the same version-checked,
+// TTL-backed caching via fetchWithVersionCache below, each keyed off its
+// own Map instance (yearStatsCache / monthStatsCache) — separate Map
+// instances per period type, rather than a shared Map keyed by group id
+// alone, so a year lookup and a month lookup for the same group can never
+// collide on the same cache entry.
+export const STATS_CACHE_TTL_MS = 60 * 60 * 1000;
+
+type StatsCacheEntry<T> = { version: number; expiresAt: number; stats: T };
+
 export const sessionStatsCache = new Map<
 	number,
-	{ version: number; expiresAt: number; stats: SessionStatsData }
+	StatsCacheEntry<SessionStatsData>
+>();
+export const yearStatsCache = new Map<
+	number,
+	StatsCacheEntry<AggregateStatsResult[] | null>
+>();
+export const monthStatsCache = new Map<
+	number,
+	StatsCacheEntry<AggregateStatsResult[] | null>
 >();
 
 export async function fetchStatsVersion(
@@ -35,12 +53,16 @@ export async function fetchStatsVersion(
 	return data?.[0]?.id ?? 0;
 }
 
-export async function fetchSessionStats(
-	viewedGroupId: number
-): Promise<SessionStatsData> {
+async function fetchWithVersionCache<T>(
+	cache: Map<number, StatsCacheEntry<T>>,
+	viewedGroupId: number,
+	fetchStats: (
+		supabase: Awaited<ReturnType<typeof getAuthenticatedSupabaseClient>>
+	) => Promise<T>
+): Promise<T> {
 	const supabase = await getAuthenticatedSupabaseClient();
 	const currentVersion = await fetchStatsVersion(supabase, viewedGroupId);
-	const cached = sessionStatsCache.get(viewedGroupId);
+	const cached = cache.get(viewedGroupId);
 	if (
 		cached &&
 		cached.version === currentVersion &&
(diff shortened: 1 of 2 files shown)


wheresrhys:
Why are you no grouping the new function by species? I expect these functions to be equivalent to fetchSessionStats except for aggregate periods

lib/underlying-stats.ts
```diff
@@ -76,3 +79,37 @@ export async function fetchSessionStats(
...
+	viewedGroupId: number
+): Promise<AggregateStatsResult[] | null> {
+	const supabase = await getAuthenticatedSupabaseClient();
+	return supabase
+		.rpc('aggregate_stats', {
+			ringing_group_filter: viewedGroupId,
+			group_by_species: false,
+			group_by_time_period: 'month'
+		})
+		.then(catchSupabaseErrors) as Promise<AggregateStatsResult[] | null>;
```
wheresrhys: These functions should reuse the same version and cache implementation as for session stats

wheresrhys:
Good catch — fixed. `fetchYearStats`/`fetchMonthStats` now call `aggregate_stats` with `group_by_species: true` (was `false`, following `fetchPayOffStats`' summary-only precedent, which is the wrong precedent here). They now return one row per period+species, matching `fetchSessionStats`' granularity (its `stats_per_day_and_species` RPC groups by day+species) — the only remaining difference is the aggregate period (year/month vs. day).

Updated the code comment above both functions and the corresponding test assertions (`group_by_species: true` in the expected RPC call args) in `lib/__tests__/underlying-stats.test.ts`. Full test suite (798 unit tests + 91 E2E) and `npm run qa` are green. Pushed as dd6e769.

wheresrhys:
Addressed in 8293dbe — `fetchYearStats`/`fetchMonthStats` now reuse `fetchSessionStats`' version-checked, TTL-backed cache instead of hitting `aggregate_stats` uncached on every call.

Factored the shared logic into a `fetchWithVersionCache` helper (still using `fetchStatsVersion` + the same TTL constant, now renamed `STATS_CACHE_TTL_MS` since it's shared) — `fetchSessionStats`, `fetchYearStats` and `fetchMonthStats` all go through it. Each keeps its own cache `Map` (`sessionStatsCache` / `yearStatsCache` / `monthStatsCache`) rather than sharing one keyed by group id alone, so a year lookup and a month lookup for the same group can never collide on the same entry.

Updated `lib/__tests__/underlying-stats.test.ts` to assert caching behaviour (cache-hit, version-bump refetch, TTL-expiry refetch, per-group isolation) for both functions, mirroring `fetchSessionStats`' existing tests, plus an explicit test that year/month caches for the same group don't collide.

https://github.com/wheresrhys/totf/pull/521#discussion_r3824230355

